PR #10, opened 2018-11-28
Feature/add s3 health check

SbiCA:
Hi @unaizorrilla 

I added S3 healthcheck. I was wondering if we should put all AWS related checks in the the namespace `AspNetCore.HealthChecks.Aws.XYZ` for instance the we could move `DynamoDb` too but I'm not sure how much this would be a breaking change?

Cheers


unaizorrilla:
Hi @SbiCA 

Let me some days to review this, probably on friday!

thanks

PR #26, opened 2018-12-13
Made TelemtryClient static for ApplicationInsights publisher

dantheman999301:
Whilst looking at the issue I raised for Service Bus, I noticed the Insights Publisher also was not kept for the lifetime of the application, so I made it only initialise once.


Xabaril:
Hi @dantheman999301 

Thanks for create this, let me this weekend to review all opened PR

unaizorrilla:
Hi @dantheman999301 

I'm working on this, and I accept this PR, but for next contributions, the PR don't contain only the code, you need to increment also the version of the modified package in dependencies.props in order to create a new nuget package.

Issue #30, opened 2018-12-14
[Question] Should URI Health Checks leverage IHttpClientFactory for management?


joncloud:
Currently the project is creating a new and disposing of `HttpClient` per health check. Microsoft introduced `IHttpClientFactory` to help abstract the lifetime management of `HttpClient` as well as some other benefits in their [original announcement](https://blogs.msdn.microsoft.com/webdev/2018/02/28/asp-net-core-2-1-preview1-introducing-httpclient-factory/). Managing `HttpClient` is also covered in a stack overflow topic: [Do HttpClient and HttpClientHandler have to be disposed?](https://stackoverflow.com/a/15708633)

CarlosLanderas:
@joncloud yes I agree it would be better. especially for HttpClientMessageHandler lifetime delegation. 
We will change it.

joncloud:
If you'd like I can try implementing this.

unaizorrilla:
PR are open!!

PR #33, opened 2018-12-14
[WIP] Updates URI Health checks to leverage IHttpClientFactory

joncloud:
Addresses concerns in #30 by adding the `Microsoft.Extensions.Http` package to include lifetime management for `HttpClient`. This also adds functional tests for the health checks.

New Dependencies:
* Runtime: `Microsoft.Extensions.Http`
* Testing: `https://httpbin.org`

I think that we can do some work to adjust the testing dependency and change it so that we integrate directly with a testing host that mimics the same functionality, or we can host the `httpbin` dependency ourselves with docker.


unaizorrilla:
Looks good, I try to merge it this weekend, thanks!

PR #39, opened 2018-12-17
Service Bus + Event Hub: Not a new connection for each check

dantheman999301:
Tried to keep as close to the Redis implementation as I could for the Service Bus one as ServiceBusConnection is exposed which helps.

However, Event Hub does not expose this so I made the client initialise in the constructor instead.

I've also updated the libraries. The Service Bus update is necessary to expose the connection so needs to be at minimum 3.X.X.

If I need to move them down to a just the latest major I can do that.

### DIFF
--- a/src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
+++ b/src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
@@ -9,8 +9,8 @@ namespace HealthChecks.AzureServiceBus
     public class AzureEventHubHealthCheck
         : IHealthCheck
     {
-        private readonly string _connectionString;
-        private readonly string _eventHubName;
+        private readonly EventHubClient _eventHubClient;
+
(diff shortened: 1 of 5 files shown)


unaizorrilla:
Thanks for contributing this @dantheman999301 , I try to review this a soon as posible!

unaizorrilla:
Hi @dantheman999301 

Merged but I perform some changes, please note for next pr:

- If you update some package you need to update the version on dependencies.props
- If you update nuget package the version are on dependencies props and csproj use $() variables to set this versions.
- All if conditions need curly brackets, it is our convention


Also, I perform some changes in your code. Queue and Topic create in each healt a QueueClient and TopicClient, you reuse only the ServiceBus connection. I change to reuse the QueueClient and TopicClient and reuse the same TCP connection. The healthcheck time is improved a lot.

At this moment the build is on app veyor, if the build work ok I tag the repo in order to generate the new package version.

Thanks
Unai

PR #40, opened 2018-12-17
Replace instantiation of HttpClient with IHttpClientFactory

roketworks:
I replaced usages of `HttpClient` with named instances created via `IHttpClientFactory`. I see this is in use in the `HealthChecks.Uris` package and I thought it would be a nice tidy-up to use this pattern in the other projects where `HttpClient` is in use. 

I think it gives the added benefit of abstracting the configuration of the clients away from the implementation. Example, I was able to remove the `KubernetesClient` and replace it with configuration extensions.


unaizorrilla:
Ji @roketworks 

Merged, with a few minor changes, the release build will create a  new package for idsvr project with version 2.2.2

PR #54, opened 2019-01-03
[Azure Service Bus] Multiple Queue on same connection string bug fix

dantheman999301:
Noticed today whilst trying out the new fixes for the re-used QueueClient that if you use the same connection string but are checking against a different queue for health it will only ever check one as it is keyed off the connection string.

I've changed the code to key off both the connection string and the name of the queue.


unaizorrilla:
Hi @dantheman999301 

Can you update dependencies.props to increment this package version?

dantheman999301:
Done 👍

PR #56, opened 2019-01-07
[Azure Service Bus] Move out string to connection key as it was not being used for adding

dantheman999301:
Sorry about this, I just realised the code I committed before was pretty flawed! Was not actually using the new key for adding so it would work once then fail on every subsequent request.

Updated `dependencies.props` this time as well


unaizorrilla:
Thanks @dantheman999301

PR #67, opened 2019-01-19
Added health check for Hangfire

pajzo:
Checking how many failed jobs the Hangfire instance currently has.


unaizorrilla:
Hi

I try yo check this pr this week, thanks

unaizorrilla:
Thanks @pajzo 

Can you add more scenarios to improve this HealthChecks? at this moment is only valid to check the number of failed jobs, but I think this can be improved with more user scenarios like check specified jobs by name of something similar.

unaizorrilla:
Ok, thanks!

Issue #68, opened 2019-01-22
HealthChecks should never throw and always return a HealthCheckResult


vertonghenb:
Currently some `HealthChecks` can `throw,` but they should always return a `HealthCheckResult`.

For example in the `AzureBlobStorage`
https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/7ee7a278279bbc776e309affc3304826905eb4f0/src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs#L16

Throws:
```
FormatException: Settings must be of the form "name=value".
Microsoft.WindowsAzure.Storage.CloudStorageAccount+<>c.<Parse>b__97_0(string err) in CloudStorageAccount.cs, line 529
```

The `CloudStorageAccount.Parse(connectionString)` throws if the `connectionstring` is invalid. A solution for this would be to take in the connectionstring in the `constructor` as we do now, but parse the string in the `CheckHealthAsync` function. Which will `catch` the exception and return a `HealthCheckResult`

vertonghenb:
Other HealthChecks that have this behavior:
https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/7ee7a278279bbc776e309affc3304826905eb4f0/src/HealthChecks.AzureStorage/AzureQueueStorageHealthCheck.cs#L16

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/7ee7a278279bbc776e309affc3304826905eb4f0/src/HealthChecks.AzureStorage/AzureTableStorageHealthCheck.cs#L16

unaizorrilla:
Hi @vertonghenb 

I try to check this issue today, thanks for fill this

vertonghenb:
Dont worry i'll open a PR today.

vertonghenb:
As an addition we could log a ValidationError from here: https://github.com/Azure/azure-storage-net/blob/master/Lib/Common/NameValidator.cs

unaizorrilla:
merged!

PR #69, opened 2019-01-22
Azure * storage throws when connectionstring is invalid

vertonghenb:
Based on the issue: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/68

Read commit per commit (4), the test is included in the commit itself.

Basically the `connectionstring` is parsed in the `CheckHealthAsync` function instead of the `constructor`.
Following HealthChecks in Scope:
- Azure Blob Storage
- Azure Queue Storage
- Azure Table Storage


unaizorrilla:
Look good, I try to check The PR this week,  thanks.

PR #73, opened 2019-01-24
Use CancellationToken and implement timeout when possible

NatMarchand:
Hey there!
I would like to add a timeout on health checks.
For this, I've used as often as possible the usage of cancellationtoken and used a cancellationtokensource to trigger the timeout.

### DIFF
--- a/src/HealthChecks.Aws.S3/DependencyInjection/S3HealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.Aws.S3/DependencyInjection/S3HealthCheckBuilderExtensions.cs
@@ -2,6 +2,7 @@
 using Microsoft.Extensions.Diagnostics.HealthChecks;
 using System;
 using System.Collections.Generic;
+using System.Threading;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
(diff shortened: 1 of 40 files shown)


unaizorrilla:
Hi @NatMarchand 

I understand this PR and the feature you propose is interesting. My main issue is about who is reponsible for cancel the check due a timeout. In my opinion the health check service need to be responsible about it, like we do on old Xabaril HealthChecks. I try to add this feature in 2.2 but without time to be released on 2,2.

I don't know if this feature is on the backlog for 3.0?  @rynowak ?

NatMarchand:
I see what you mean. I can move my cancellationtoken around this line:
https://github.com/aspnet/Extensions/blob/d66169865dd078ef4877f128968d50786bbbe06f/src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs#L75

However, I think that it's interesting to have different timeout values for each check (for example I expect my db to fininsh within 20ms, whereas I expect a http call to finish within 200ms). We could add a property on HealtcheckEntry to configure the timeout for the entry.

Please note also that if this PR isn't merged, I might make another one with few things I've caught along the way (missing dispose, usage of the request cancellationtoken whenever possible, etc.)

unaizorrilla:
"Please note also that if this PR isn't merged, I might make another one with few things I've caught along the way (missing dispose, usage of the request cancellationtoken whenever possible, etc.)" 
 -> Perfect

rynowak:
@unaizorrilla - the current status is here: https://github.com/aspnet/AspNetCore/issues/2577#issuecomment-429454564

We haven't made any real plans around this issue because it's come up 1-2 times.

JoshKeegan:
This has been done for .NET Core 3 https://github.com/aspnet/AspNetCore/issues/2577#issuecomment-488712702

PR #83, opened 2019-02-05
Fixed 3 static class name.

suadev:
A minor fix to fix 3 static class name


CarlosLanderas:
@unaizorrilla I'm gonna merge this, but I do not increase package versions as this are internal aesthetic changes.

Thanks for your contribution @suadev

PR #91, opened 2019-02-08
Save detailed report for Application Insights Publisher

lfraile:
Saves (optionally) detailed event for each HealthCheck report on Application Insights. It is configured with the _saveDetailedReport_ optional parameter in _AddApplicationInsightsPublisher_.

If set to true it will save an event for each HealthCheck report.

The default value is false.


unaizorrilla:
Thanks @lfraile

PR #102, opened 2019-02-18
extended rabbitMq health check with tls option.

hansehe:
The rabbitMq health check was missing tls option, thus it would be useless in many cases where tls with rabbitMq is enabled.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -71,7 +71,7 @@
     <HealthCheckSqlite>2.2.0</HealthCheckSqlite>
     <HealthCheckUri>2.2.2</HealthCheckUri>
     <HealthCheckRedis>2.2.2</HealthCheckRedis>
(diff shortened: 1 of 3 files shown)


unaizorrilla:
Hi @hansehe 

The changes contains new parameter on RabbitMQHealthCheck but the extension method for registering the health check don't use this new parameter.

How the user can set the SSL option when the healthcheck is registered?

Also, increment the version for this package on dependencies.props

unaizorrilla:
Hi @hansehe 

Can you address the changes for this PR?

hansehe:
Hi, and I'm sorry for being so slow to reply. I missed the extensions class and the version update, but it should all be fine now. Please have a look at the update, and tell me if something should be changed.

unaizorrilla:
Thanks @hansehe for contribute this, merged and release build enqueued, when finish a new packages version (2.2.1) will be on nuget.

PR #104, opened 2019-02-18
Issue 101 - Fix multiple webhook notifications

rob-baldwin:


unaizorrilla:
Hi @rob-baldwin 
Thanks for contributing this, I increment the version on UI to 2.2.16, when release build finish a new package will be on nuget.org

Issue #122, opened 2019-03-05
OutOfMemoryException when on app service


youngcm2:
After deploying the appservice starts throwing this error. I have to restart the app to get it back up.

Formatting stinks so I attached screen shots.

**Startup.cs**

**HealthChecks**

`
services.AddHealthChecks()
                .AddSqlServer(_configuration["sql:connectionString"])
                .AddMongoDb(mongodbConnectionString: _configuration["mongo:connectionString"],
                    mongoDatabaseName: _configuration["mongo:databaseName"])
                .AddRedis(_configuration["redis:connectionString"])
                .AddAzureBlobStorage(_configuration["blobStorageOptions:uri"])
                .AddHangfire(options =>
                {
                    options.MinimumAvailableServers = 1;
                })
              .AddApplicationInsightsPublisher(_configuration["ApplicationInsights:InstrumentationKey"]);
`

**Middleware**

`
builder.UseHealthChecks("/health-check", new HealthCheckOptions
            {
                Predicate = _=>true,
                ResponseWriter = UIResponseWriter.WriteHealthCheckUIResponse
            });
`

**StackTrace**

`
Health check "HealthChecks.Publisher.ApplicationInsights.ApplicationInsightsPublisher" threw an unhandled exception after 20.4038ms
System.Threading.Tasks.TaskSchedulerException: An exception was thrown by a TaskScheduler. ---> System.OutOfMemoryException: Exception of type 'System.OutOfMemoryException' was thrown.
   at System.Threading.Thread.StartInternal()
   at System.Threading.Thread.Start()
   at System.Threading.Thread.Start(Object parameter)
   at System.Threading.Tasks.ThreadPoolTaskScheduler.QueueTask(Task task)
   at System.Threading.Tasks.TaskScheduler.InternalQueueTask(Task task)
   at System.Threading.Tasks.Task.ScheduleAndStart(Boolean needsProtection)
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ScheduleAndStart(Boolean needsProtection)
   at System.Threading.Tasks.Task.InternalStartNew(Task creatingTask, Delegate action, Object state, CancellationToken cancellationToken, TaskScheduler scheduler, TaskCreationOptions options, InternalTaskOptions internalOptions)
   at Microsoft.ApplicationInsights.Channel.InMemoryTransmitter..ctor(TelemetryBuffer buffer)
[...]

PR #123, opened 2019-03-05
Cleaned old code for ApplicationInsightsPublisher Issue #122

lfraile:
It created non-used and non-disposed objects for TelemetryConfiguration which can cause OutOfMemory exceptions. Cleaned the old code ensureing only one TelemetryConfiguration object is created for a singleton instance of TelemetryClient


unaizorrilla:
Thanks @lfraile

PR #130, opened 2019-03-20
Add support for custom tags and failure status of azure storage queues

Marcel-Lambacher:
# General

In our project, we've faced an issue regarding the health checks of the storage queue within the package: **AspNetCore.HealthChecks.AzureStorage**

The problem was, that no matter what connection string has been passed to the health check, the result was always healthy.
After a quick look in the code, we found out, that the handling of the storage queue doesn't follow the same behavior as the tables and blobs.

For all three storage types, there is an option to pass user defined tags to the health check. However, the health check extension for the queue ignores them and uses an internal identifier.

# Solution

The simple solution here is to align the behavior of the storage queue to the behavior of the table and blobs.

# What has been changed

Instead of `new string[] { AZUREQUEUE_NAME }` which has been used as tags, no matter what the user has been passed to the extension method, the tags from the parameter will be passed to the health check builder now.

Corresponding unit tests has been added too.

### DIFF
--- a/src/HealthChecks.AzureStorage/DependencyInjection/AzureStorageHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.AzureStorage/DependencyInjection/AzureStorageHealthCheckBuilderExtensions.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Extensions.DependencyInjection
         /// the default status of <see cref="HealthStatus.Unhealthy"/> will be reported.
         /// </param>
         /// <param name="tags">A list of tags that can be used to filter sets of health checks. Optional.</param>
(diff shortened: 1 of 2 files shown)


unaizorrilla:
Hi Marcel,

Thanks for fill the issue, I'm on MVP summit, let me a few days to check and review this PR.

unaizorrilla:
Hi @Marcel-Lambacher ,

First of all, thanks for contribute this and your clear explanation of the issue !!

Merged on master, a new release build is on appveyor, when finish a new healthcheck package for Azure Storage will be on NuGet with version 2.2.2!

thanks 
Unai

PR #139, opened 2019-04-02
Added overload to Raven Health check to allow user to pass RavenDBOptions

ankitvijay:
With this pull request, I have added a method overload to allow user to pass <s>array of Raven URLs</s> `RavenDBOptions` instead of just single connection string. Note that existing method which allows the user to pass single connection string is still there but it has been marked `Obsolete`.

**Why this change?**
- This change is useful to test the health check for the Raven DB Cluster (that is multiple nodes) instead of a single connection string. For Load Balance and Fail over scenarios we would have multiple connection strings instead of one.
- In addition to this, `RavenDBOptions` also supports certificate authentication. That is user can _optionally_ pass the authentication certificate, which should ideally be a requirement for any production workload (like ours)

As an alternate (without the changes in this PR), we can add multiple raven DB urls as pseudocode below:

        var counter = 1;
        foreach(var url in urls)
        {
            healthCheckBuilder.AddRavenDB(url, "DemoDbName",$"ravendb node{counter}, url: {url} ");
            ++counter;
        }

However, this is less than ideal for following reasons:
- It means additional lines of code.
- If one of urls is down/ unavailable the overall health status would be returned as "Unhealthy" which is incorrect since server can still connect to the alternate Raven DB url.
- In scenarios where authentication is required we cannot use this health check library.

### DIFF
--- a/src/HealthChecks.RavenDB/DependencyInjection/RavenDBHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.RavenDB/DependencyInjection/RavenDBHealthCheckBuilderExtensions.cs
@@ -1,4 +1,5 @@
-﻿using HealthChecks.RavenDB;
+﻿using System;
+using HealthChecks.RavenDB;
 using Microsoft.Extensions.Diagnostics.HealthChecks;
 using System.Collections.Generic;
 
@@ -22,7 +23,8 @@ namespace Microsoft.Extensions.DependencyInjection
         /// the default status of <see cref="HealthStatus.Unhealthy"/> will be reported.
         /// </param>
         /// <param name="tags">A list of tags that can be used to filter sets of health checks. Optional.</param>
-        /// <returns>The <see cref="IHealthChecksBuilder"/>.</returns></param>
+        /// <returns>The <see cref="IHealthChecksBuilder"/>.</returns>
+        [Obsolete("This method is deprecated.")]
         public static IHealthChecksBuilder AddRavenDB(
             this IHealthChecksBuilder builder,
             string connectionString,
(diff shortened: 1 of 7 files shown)


unaizorrilla:
Hi @ankitvijay 

I try to review this ASAP!

src/HealthChecks.RavenDB/RavenDBHealthCheck.cs
```diff
@@ -11,20 +11,33 @@ namespace HealthChecks.RavenDB
...
         private readonly string _connectionString;
+        private readonly string[] _urls;
         private readonly string _specifiedDatabase;
 
         public RavenDBHealthCheck(string connectionString, string databaseName = default)
         {
             _connectionString = connectionString ?? throw new ArgumentNullException(nameof(connectionString));
             _specifiedDatabase = databaseName;
         }
+
```
unaizorrilla: Hi, probably you can use only one constructor for both cases!

test/FunctionalTests/HealthChecks.RavenDB/RavenDBWithSingleConnectionStringHealthCheckTests.cs
```diff
@@ -13,12 +13,12 @@
 namespace FunctionalTests.HealthChecks.RavenDB
 {
     [Collection("execution")]
-    public class ravendb_healthcheck_should
+    public class ravendb_with_single_connection_string_healthcheck_should
```
unaizorrilla: Well, if you use only one constructor, probably create tests for one or more connectionstring doens't have sense, true?

src/HealthChecks.RavenDB/DependencyInjection/RavenDBHealthCheckBuilderExtensions.cs
```diff
@@ -35,5 +35,34 @@ public static class RavenDBHealthCheckBuilderExtensions
...
+        /// </param>
+        /// <param name="failureStatus">
+        /// The <see cref="HealthStatus"/> that should be reported when the health check fails. Optional. If <see langword="null"/> then
+        /// the default status of <see cref="HealthStatus.Unhealthy"/> will be reported.
+        /// </param>
+        /// <param name="tags">A list of tags that can be used to filter sets of health checks. Optional.</param>
+        /// <returns>The <see cref="IHealthChecksBuilder"/>.</returns></param>
+        public static IHealthChecksBuilder AddRavenDB(
+            this IHealthChecksBuilder builder,
+            string[] urls,
```
unaizorrilla: Could be "connectionsstrings" better name than "urls"? or "url" instead of connectionstring in the other overload!

src/HealthChecks.RavenDB/RavenDBHealthCheck.cs
ankitvijay: Correct. I can fix this.

src/HealthChecks.RavenDB/DependencyInjection/RavenDBHealthCheckBuilderExtensions.cs
ankitvijay: Hi @unaizorrilla I did consider using `connectionStrings`. However, decided aginast it since the raven documentation uses the term `url` instead of connectionString. I actually thought to rename original method parameter to `url` instead, but since it is an implementation already used by consumers did not want to change it.

ankitvijay: I had an additional suggestion if you like to consider it. It is to make the original method as `Obsolete` since Raven takes url array in `DocumentStore`. This way we would be consistent with the Raven documentation/implementation. However, will leave the decision on you..

unaizorrilla: Ok!

ankitvijay: Great will make the relevant changes and update the PR.

unaizorrilla: Thansk 

I try to merge asap!

unaizorrilla: Please let me know when you update the PR

src/HealthChecks.RavenDB/RavenDBHealthCheck.cs
ankitvijay: Done.

test/FunctionalTests/HealthChecks.RavenDB/RavenDBWithSingleConnectionStringHealthCheckTests.cs
ankitvijay: HI @unaizorrilla I realized that we would still need the separate tests since we still have two method overloads in `RavenDBHealthCheckBuilderExtensions` even though the constructor is one.

src/HealthChecks.RavenDB/DependencyInjection/RavenDBHealthCheckBuilderExtensions.cs
ankitvijay: Hi @unaizorrilla I have made some additional changes to allow user to pass raven auth certificate. To achieve this I have introduced a new parameter `RavenDBOptions` much inspired by existing code for other providers. In addition to this, the existing method which takes single `connectionString` as overload has been marked `Obsolete`

ankitvijay:
Hi @unaizorrilla when you get a chance could you look into my PR please. I'm waiting for it to merge to start using it my project. Thanks!

unaizorrilla:
Yeap, tomorrow!

unaizorrilla:
Merged, thanks for contributing. Build Release is on appveyor, when finish a new package 2.2.2. for RavenDb will be on NuGet.

ankitvijay:
Thanks @unaizorrilla

PR #140, opened 2019-04-03
Add cosmos db support

Ahmad-Magdy:
Adding CosmosDb support, because documentdb will be deprecated.
I followed the same flow of documentdb and didn't delete it, so the maintainers can remove documentDb when it's officially deprecated.
I used ConnectionString instead of Uri and access token because i think it's simpler and easier to use.


build/dependencies.props
```diff
@@ -84,6 +85,7 @@
     <HealthCheckSystem>2.2.1</HealthCheckSystem>
     <HealthCheckNetwork>2.2.2</HealthCheckNetwork>
     <HealthCheckDynamoDb>2.2.0</HealthCheckDynamoDb>
+    <HealthCheckCosmosDb>2.2.0</HealthCheckCosmosDb>
```
unaizorrilla: If this package use preview packages, set also as preview version, 2.2.0-preview!

Ahmad-Magdy:
@unaizorrilla Great!
Done.

Issue #138, opened 2019-04-01
Create option to publish only healthreports reporing a failure


Expecho:
Given the `AddApplicationInsightsPublisher()` method, I would love to have it support a flag indicating only reports indicating a failure are send to Application Insights. I am only interested in the amount of failures over time and since I am paying for the amount of data ingested I would like to filter out the normal state.

Other option would be to create an `ITelemetryProcessor` to filter out the items but I'd rather have built-in support.

I am more than willing to create a PR so we can take a closer look at how an implementation might look like.

unaizorrilla:
You can filter on Application Insights using the status property if you are interested only on failures.

Expecho:
I understand, but I do pay for that ingested data. So I would prefer to prevent it from entering Application Insights at all.

unaizorrilla:
You can reduce the amount of data on AI using sampling!

Expecho:
Also correct, but that might sample out unhealthy reports as well.

Expecho:
So summarized, this are the current options:

- Sampling, but the drawback here is that unhealthy reports are excluded as well
- Telemetry Filter, best option for now but requires more work
- Filter in the query using the portal, but the drawback here is we still pay for the data we are not interested in and we send data over the wire we do not need/want

So I still see added value for being able to create a flag indicating only reports indicating a failure are send to Application Insights.

unaizorrilla:
I'm not sure, let me check this with more people on the project and I try to respond you ASAP!

lfraile:
After talking with @unaizorrilla we think waht you propose, @Expecho, makes sense.

Let me a couple of days to work in it.

unaizorrilla:
or @Expecho you can contribute this?

Expecho:
Sure, no problem. I will give it a go later this week!

Expecho:
@lfraile @unaizorrilla I think we need to discuss the testing part of this proposed change. See the PR for details.

lfraile:
Will take a look tomorrow

________________________________
De: Peter Bons <notifications@github.com>
Enviado: Thursday, April 4, 2019 9:13:13 PM
Para: Xabaril/AspNetCore.Diagnostics.HealthChecks
Cc: Luis Fraile Hernandez; Mention
Asunto: Re: [Xabaril/AspNetCore.Diagnostics.HealthChecks] Create option to publish only healthreports reporing a failure (#138)


[...]

PR #143, opened 2019-04-04
exclude reports that indicate a healthy status

Expecho:
Fix for #138 that makes it possible to configure the Application Insights publisher to only publish reports that indicate an unhealthy state by adding an optional flag `excludeHealthyReports`. 

When the publisher is configured to publish the details (using `saveDetailedReport`) only those detailed reports indicating an unhealthy state will be published if `excludeHealthyReports` is set to `true`

I am all open for creating a functional test for this (as there are none for this particalur publisher) but I would like to add a dependency to [Moq](https://github.com/Moq/moq4/wiki/Quickstart) or [NSubstitute](https://nsubstitute.github.io/) in order to be able to fake the application insights classes involved. If that is ok with you I can add those tests, otherwise I need to figure out other ways.


unaizorrilla:
Hi @Expecho ,

Thanks for contributing this, a new release build is on appveyor, when finish a new package for AI publisher version 2.2.3 will be on NuGet!

Issue #29, opened 2018-12-13
Immutable Deployments - Option for in memory (or other alternative) healthcheckdb for UI?


cscheib:
Have you considered options other than persisting a sqlite file to disk for the healthcheckdb?

We've been moving toward an immutable runtime environment model (Azure App Service using "Run From Package"), which means writing that file to disk is going to fail.

It looks like I can forgo instantiating the UI, and just query /healthchecks-api for check results, which may suffice for our needs.  But, a UI is always nice.

CarlosLanderas:
@unaizorrilla what do you think about this. It's obvious that the UI is not prepared for this kind of scenario. Could we implement additional providers?

unaizorrilla:
Well, this is not a priority on this moment!

unaizorrilla:
Hi @cscheib 

I need to create a new feature for supporting specifying the database path, probably I try to add support for sqlite on memory also on this new feature.

HalidCisse:
I had same issue before when my app fail to start because it cant write to the disk, how about an efcore dbcontext implementation ?

unaizorrilla:
Hi @HalidCisse 

With new versions you can specify the db path for Sqlite. Check UIAndApi sample and the new appsettings.json parameter configuration.

Next version of the ui probably include a memory database support!

SychevIgor:
Totally agree with @cscheib , we need at least in memory provider or better - they way to configure persistence. 
As a work-around- we can replace the call AddHealthChecksUI from an library
https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.UI/ServiceCollectionExtensions.cs
without own code, where could by copy-pasted an original call, but without SQLLite

```
 services.AddDbContext<HealthChecksDb>(db =>
            {
                var connectionString = healthCheckSettings.HealthCheckDatabaseConnectionString;
                if (string.IsNullOrWhiteSpace(connectionString))
                {
                    var contentRoot = configuration[HostDefaults.ContentRootKey];
                    var path = Path.Combine(contentRoot, databaseName);
                    connectionString = $"Data Source={path}";
                }
                else
                {
                    connectionString = Environment.ExpandEnvironmentVariables(connectionString);
                }
                db.UseSqlite(connectionString);
            });
```

unaizorrilla:
Hi @SychevIgor 

We try to add this feature on next ui version, thanks for your feedback.

Unai

[...]

PR #148, opened 2019-04-10
Add ability to select data provider for health checks ui

Odonno:
## Description

Related to #29 

Add ability to select the data provider used by the health checks ui, among : 

* In Memory
* Sqlite
* SQL Server

## TODO

* [x] Use the data provider selected in the appsettings
* [x] Update the configuration of the appsettings
* [x] Unit tests
* [ ] readme documentation

## Breaking changes

It contains breaking changes : 

* the in-memory store is used by default (if no configuration)
* the previous appsettings configuration has been replaced by a `DataSettings` object to let devs select the data provider they want


unaizorrilla:
Hi @Odonno 

We are working on next version for the UI. This is not a easy feature for some reasons:

 - EF InMemory is not a production ready provider.
 - Change from persisted to inmemory Sqlite also has some effects on the lifecicle of context, with memory this need to be singleton.

We try to address this on next version, if you want contribute please do it on vnext branch.

Odonno:
Hi @unaizorrilla 

I did not know that EF Core InMemory was not production ready. I know it was used for testing purpose but I don't see why it cannot be used in production.

Either way, if it cannot be used, we have to conditionally load the EF context and create an in-memory object by default. But that seems like a huge change.

Issue #173, opened 2019-05-02
Register connection string from factory


JoshKeegan:
Hi,
I've just started to use this library and it's working great :)  

I'm using it for SQL Server, and we keep our connection details in an object that gets wired up via dependency injection. This means at the point we come to register the health check, the `IServiceProvider` hasn't been built, so we can't get at it. This means I can't use your `pservices.AddSqlServer(...` method (which requires the connection string) and instead must use `services.Add(new HealthCheckRegistration(...` from Microsoft.Extensions.Diagnostics.HealthChecks because it can accept a factory to build the health check (allowing us to access an `IServiceProvider`).

Would there be any interest in a new `services.AddSqlServer(...` method that supports a factory for the connection string?
The same thing could be done for the other health checks if there's interest?

### Existing usage
```cs
services.AddHealthChecks()
	.Add(new HealthCheckRegistration("db", provider =>
	{
		DatabaseSettings dbSettings = provider.GetService<DatabaseSettings>();
		return new SqlServerHealthCheck(dbSettings.ConnectionString, "SELECT 1;");
	}, null, new string[] { "readiness" }));
```

### Proposed usage
If there was a factory for building the connection string from the `IServiceProvider`, this could become:
```cs
services.AddHealthChecks()
	.AddSqlServer(provider => provider.GetService<DatabaseSettings>().ConnectionString,
		name: "db", tags: new string[] {"readiness"});
```

unaizorrilla:
Hi @JoshKeegan 

Sound interesting !  Can you send a PR? and we can start discussing implementation details!

JoshKeegan:
@unaizorrilla Will do :)

JoshKeegan:
@unaizorrilla Any thoughts on the PR?

CarlosLanderas:
Merged.

PR #175, opened 2019-05-02
Register a SQL Server Health Check with a factory for the connection string

JoshKeegan:
As discussed in #173 

This allows you to use `services.AddSqlServer(...` if you use dependency injection for your configuration. Sample usage:
```cs
services.AddHealthChecks()
	.AddSqlServer(provider => provider.GetService<DatabaseSettings>().ConnectionString,
		name: "db", tags: new string[] {"readiness"});
```

which would currently require using `services.Add(new HealthCheckRegistration(...` from Microsoft.Extensions.Diagnostics.HealthChecks:
```cs
services.AddHealthChecks()
	.Add(new HealthCheckRegistration("db", provider =>
	{
		DatabaseSettings dbSettings = provider.GetService<DatabaseSettings>();
		return new SqlServerHealthCheck(dbSettings.ConnectionString, "SELECT 1;");
	}, null, new string[] { "readiness" }));
```

If this seems reasonable for SQL Server, it may make sense for most (all?) other health checks too.

### DIFF
--- a/src/HealthChecks.SqlServer/DependencyInjection/SqlServerHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.SqlServer/DependencyInjection/SqlServerHealthCheckBuilderExtensions.cs
@@ -1,4 +1,5 @@
-﻿using HealthChecks.SqlServer;
+﻿using System;
+using HealthChecks.SqlServer;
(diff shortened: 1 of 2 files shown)


CarlosLanderas:
@JoshKeegan  I've been testing your PR and it looks good

Issue #172, opened 2019-05-01
DocumentDb HealthCheck new-ing up DocumentClient per call


ymitis:
The [DocumentDbHealthCheck](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.DocumentDb/DocumentDbHealthCheck.cs) creates a new instance of DocumentClient for each call to CheckHealthAsync.

[Official Documentation](https://azure.microsoft.com/en-us/blog/performance-tips-for-azure-documentdb-part-1-2/) suggests using a Single instance of the DocumentClient per lifetime of the application.

`

            try
            { 
                using (var documentDbClient = new DocumentClient(
                    new Uri(_documentDbOptions.UriEndpoint),
                    _documentDbOptions.PrimaryKey))
                {
                    await documentDbClient.OpenAsync(cancellationToken);
                    return HealthCheckResult.Healthy();
                }
            }
            catch (Exception ex)
            {
                return new HealthCheckResult(context.Registration.FailureStatus, exception: ex);
            }
`

unaizorrilla:
We fix this issue on other healtchecks, I try to fix it ASAP.

unaizorrilla:
Fixed on 2.2.1 package, this will be available when appveyor finish the release build!

PR #176, opened 2019-05-02
Prevent CosmosDbHealthCheck form instantiating a new CosmosClient per call

ymitis:
This is similar to the recent fix for Issue [172](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/172).


unaizorrilla:
Merged, thanks again for contribute into this healthcheck!

PR #178, opened 2019-05-03
Fix functional tests running locally

JoshKeegan:
wnameless/oracle-xe-11g docker image no longer exists as Oracle DMCA'd it. See https://github.com/wnameless/docker-oracle-xe-11g/issues/118  

This PR changes it to be built from source instead.  

Note that if you already have the wnameless/oracle-xe-11g docker image cached locally, it will still be working for you, so delete that image to reproduce the issue.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -62,7 +62,9 @@ services:
         ports:
           - ${IDSVR_PORT}:80
     oracle:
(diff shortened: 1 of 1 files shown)


unaizorrilla:
Merged, I also added ravendb as container to improve reliability of functional tests with Raven!

Thanks for contribute!

PR #184, opened 2019-05-16
Feature/named http client

kogratte:
I add the possibility to use a named httpClient (with a different name than the urlGroup)


kogratte:
Seams to be buggy. Need more checks.

PR #202, opened 2019-06-08
Fix concurrency bug when two requests are sent at the same time

ittennull:
This PR fixes concurrency issue.
Also in general it's preferable to use standard methods that were properly tested than trying to repeat them for no reason

Can I ask you to merge it asap because this is a real issue in production and I had to go back to old version of the package that doesn't have the bug


unaizorrilla:
Hi @ittennull 

Let me test it, If works I try to merge it ASAP

unaizorrilla:
I'm sorry for delay on review this, I do this week!!

:-(

unaizorrilla:
Hi @ittennull 

Merged, thanks again for contribute with this PR!!

Issue #239, opened 2019-07-22
Add Datadog publisher


rpopovych:
Datadog is a popular metrics and event aggregator that is often used to help operate and maintain your infrastructure. One of its features is service checks, which is very similar to health checks. I would like to add a publisher that pushes the health check results into Datadog as custom service checks.

Datadog Service Checks: https://docs.datadoghq.com/monitors/guide/visualize-your-service-check-in-the-datadog-ui/#pagetitle

rpopovych:
I will make a PR shortly with the implementation.

unaizorrilla:
Thanks @rpopovych 

The PR is on review, please, try to address the feedback!

Thanks for contribute!

renatogbp:
Hi,

I have being trying to use the DataDog publisher, but it isn't publishing any metrics to datadog. 
This is my Setup:
```
var hcBuilder = services.AddHealthChecks();
hcBuilder.AddDatadogPublisher("serviceName.healthchecks");
```
Can someone add tests to make sure this is working, please?

I have been using DataDog for other metrics and that's how I configure:
```
public static void Configure(string serverName = null, int? port = null, string prefix = null)
{
	var dogstatsdConfig = new StatsdConfig();

	if (!string.IsNullOrWhiteSpace(serverName))
	{
		dogstatsdConfig.StatsdServerName = serverName;
	}

	if (port.HasValue)
	{
		dogstatsdConfig.StatsdPort = port.Value;
	}

	if (!string.IsNullOrWhiteSpace(prefix))
	{
		dogstatsdConfig.Prefix = prefix;
	}

	DogStatsd.Configure(dogstatsdConfig);
}
```

rpopovych:
Sorry, I missed this message. The extension method configures it's own Datadog service internally, and currently exposes a parameter to configure the Datadog agent name.

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/5dde843c51c7a4b6778c444adae6a95527f58068/src/HealthChecks.Publisher.Datadog/DependencyInjection/DatadogHealthCheckBuilderExtensions.cs#L18

Using the `IDogStatsd` interface allows us to test more easily. I left a question on the PR about the correct place for the publisher tests, but did not get a response: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/240#discussion_r306427500

[...]

PR #240, opened 2019-07-23
Datadog Publisher

rpopovych:
Implementing #239: Creating a new Datadog publisher package.

The new package uses the official Datadog DogStatsD library: https://github.com/DataDog/dogstatsd-csharp-client.
The recommended Datadog strategy is to send custom metrics to the agent and let the agent handle communication with Datadog APIs.
The assumption here is that the datadog agent is running locally, either as a process or a sidecar container. If that's not the case, agent server name is configurable.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -97,6 +97,7 @@
     <HealthCheckUI>2.2.32</HealthCheckUI>
     <HealthCheckUIClient>2.2.4</HealthCheckUIClient>
     <HealthCheckPublisherAppplicationInsights>2.2.4</HealthCheckPublisherAppplicationInsights>
+    <HealthCheckPublisherDatadog>2.2.0</HealthCheckPublisherDatadog>
     <HealthCheckPublisherPrometheus>2.2.0</HealthCheckPublisherPrometheus>
     <HealthCheckAWSS3>2.2.0</HealthCheckAWSS3>
     <HealthCheckKeyVault>2.2.2</HealthCheckKeyVault>
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -64,6 +64,7 @@
     <KubernetesClient>1.2.0</KubernetesClient>
     <MicrosoftAspNetCoreSignalRClient>1.1.0</MicrosoftAspNetCoreSignalRClient>
     <MicrosoftAspNetCoreSignalR>1.1.0</MicrosoftAspNetCoreSignalR>
+    <DogStatsDCSharpClient>3.3.0</DogStatsDCSharpClient>
   </PropertyGroup>
 
   <PropertyGroup Label="CLI Tools Versions">
--- a/src/HealthChecks.Publisher.Datadog/DatadogPublisher.cs
+++ b/src/HealthChecks.Publisher.Datadog/DatadogPublisher.cs
@@ -1,0 +1,64 @@
+﻿using Microsoft.Extensions.Diagnostics.HealthChecks;
+using StatsdClient;
+using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace HealthChecks.Publisher.Datadog
+{
+    class DatadogPublisher
+        : IHealthCheckPublisher
+    {
+        private readonly IDogStatsd _dogStatsd;
+        private readonly string _serviceCheckName;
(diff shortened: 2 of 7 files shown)


build/dependencies.props
```diff
@@ -97,6 +97,7 @@
     <HealthCheckUI>2.2.32</HealthCheckUI>
     <HealthCheckUIClient>2.2.4</HealthCheckUIClient>
     <HealthCheckPublisherAppplicationInsights>2.2.4</HealthCheckPublisherAppplicationInsights>
+    <HealthCheckPublisherDatadog>2.2.0</HealthCheckPublisherDatadog>
```
rpopovych: I wasn't sure what the appropriate versioning is for this repo, but it looked like everything was at least 2.2.

src/HealthChecks.Publisher.Datadog/DatadogPublisher.cs
```diff
@@ -0,0 +1,64 @@
+﻿using Microsoft.Extensions.Diagnostics.HealthChecks;
+using StatsdClient;
+using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace HealthChecks.Publisher.Datadog
+{
+    class DatadogPublisher
```
rpopovych: I would like to add some unit tests for this class, but it looks like we only DependencyInjection and UI tests in the UnitTests project. What would be the best place to put these?
@unaizorrilla, @lurumad, @carloslanderas, @eiximenis

src/HealthChecks.Publisher.Datadog/HealthChecks.Publisher.Datadog.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <RepositoryUrl>$(RepositoryUrl)</RepositoryUrl>
+    <Company>$(Company)</Company>
+    <Authors>$(Authors)</Authors>
+    <PackageId>AspNetCore.HealthChecks.Publisher.Datadog</PackageId>
+    <PublishRepositoryUrl>$(PublishRepositoryUrl)</PublishRepositoryUrl>
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder)</AllowedOutputExtensionsInPackageBuildOutputFolder>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="DogStatsD-CSharp-Client" Version="3.3.0" />
```
rpopovych: Does this dependency need to be added to dependencies.json? Not too familiar with that file and not sure what uses it.
@unaizorrilla, @lurumad, @carloslanderas, @eiximenis

unaizorrilla:
Hi @rpopovych 

I try to review the PR ASAP, thanks for summiting this and contribute into this project!

src/HealthChecks.Publisher.Datadog/DatadogPublisher.cs
```diff
@@ -0,0 +1,64 @@
...
+                        break;
+                    case HealthStatus.Unhealthy:
+                        dataDogStatus = Status.CRITICAL;
+                        break;
+                    default:
+                        dataDogStatus = Status.UNKNOWN;
+                        break;
+                }
+
+                var tags = (_defaultTags ?? new string[0]).Concat(new[]
```
unaizorrilla: _defaultTags is always not null because you check this on constructor. Probably will be null and you need to remove this constraint on constructor!!

src/HealthChecks.Publisher.Datadog/HealthChecks.Publisher.Datadog.csproj
unaizorrilla: Yeap, please add this version ( DotStatsD-csharp-client) also in dependencies.props!

src/HealthChecks.Publisher.Datadog/DatadogPublisher.cs
rpopovych: Great catch! I'll remove the null check

rpopovych:
Thanks @unaizorrilla!
I addressed the feedback.
I have just one more question remaining about unit tests above.

unaizorrilla:
Thanks for contribute this @rpopovych

PR #242, opened 2019-07-24
Add EventStore health check using connection string. 

MuadDib:
Clarify that existing check uses URI.

### DIFF
--- a/src/HealthChecks.EventStore/DependencyInjection/EventStoreHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.EventStore/DependencyInjection/EventStoreHealthCheckBuilderExtensions.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Extensions.DependencyInjection
         /// Add a health check for EventStore services.
         /// </summary>
         /// <param name="builder">The <see cref="IHealthChecksBuilder"/>.</param>
-        /// <param name="eventStoreConnection">The EventStore connection string to be used.</param>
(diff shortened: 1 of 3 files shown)


unaizorrilla:
Hi @MuadDib 

Why not only one HealthCheck? You can create a connection string using uri /login / pass also and use the same healthcheck?

MuadDib:
Heya Unai,

Thanks for your feedback.

My thought process was:
- I don't wanna change any of the existing logic (thus not altering the existing EventStoreHealthCheck class)
- I didn't want to parse connectionString manually (to extract URI/login details) as I want to future proof it if EventStore client ever changes the format
- There is already Create method on EventStore client that takes connectionString
- Follow the same code conventions that are already in place (e.g. AzureServiceBus already has multiple health checks)

Let me know if you would like me to explain myself in more details or if I misunderstood something.

Best,
Igor

unaizorrilla:
Hi @MuadDib 

- AzureServiceBus exist different health checks for different purposes (queue, topics etc) in this case both are for the same "element" EventStore.
- You can mantain back compatibility with new registration methods but with only one health check.

Sounds ok for you?

sungam3r:
@MuadDib  Please rebase onto `master` if you are going to continue work on this PR.

sungam3r:
ping @MuadDib

sungam3r:
@NielsPilgaard are you interested in?

NielsPilgaard:
Version 6.0.2 already uses connection string. If I'm reading this issue correctly, it can be closed as completed.

sungam3r:
Thanks.

PR #266, opened 2019-09-06
Allow Service Bus health checks to be registered with empty connection strings &/or entity path

JoshKeegan:
This PR comes after an issue we have faced where a service was deployed without the connection string for a queue being supplied in the configuration.

When a health check endpoint was being called, the registration for the queue's health check was being run and throwing an exception as the connection string was `""` and the health check's ctor didn't allow this.

This is not what I would expect, I would expect registration to work and the registered health check to run but return that the queue is unhealthy.

I have looked at a few other resources health checks (e.g. SQL Server & Cosmos DB) and they do not check in the ctor for `""`, but only for `null`. So this PR changes the Service Bus behaviour to match that.

### DIFF
--- a/src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
+++ b/src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
@@ -15,21 +15,27 @@ namespace HealthChecks.AzureServiceBus
         private readonly string _eventHubConnectionString;
         public AzureEventHubHealthCheck(string connectionString, string eventHubName)
         {
-            if (string.IsNullOrEmpty(connectionString))
+            if (connectionString == null)
             {
                 throw new ArgumentNullException(nameof(connectionString));
             }
-
-            if (string.IsNullOrEmpty(eventHubName))
+            if (eventHubName == null)
             {
                 throw new ArgumentNullException(nameof(eventHubName));
             }
 
-            var builder = new EventHubsConnectionStringBuilder(connectionString)
+            if (connectionString != "" && eventHubName != "")
             {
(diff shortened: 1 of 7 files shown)


unaizorrilla:
Hi @JoshKeegan 

I try to review this asap, before migrate master to .NET Core 3.0!

unaizorrilla:
Hi @JoshKeegan 

Setting a empty connection string is a normal behavior for you? If throw, the configuration issue will appair early  but with empty connection string you know to ensure the connection string is well configured or is a fail status of the queue.. I'm not sure this behavior is correct

JoshKeegan:
Hi,
Thanks for looking at this. 

Throwing may make the issue appear earlier (during DI registration), however in the case I had the queue was not mission-critical, so we don't want the queue being unhealthy to cause the entire application to break (but we still want to know about it via health check so that we know the service is degraded). 

If the registration of a health check fails, it effectively means we cannot run our health check endpoints, so the entire app goes offline (as we use kubernetes, it's checking for liveness and readiness health checks before sending traffic to the pod). If the health check can be registered but will return Unhealthy when checked then the consumer decides what should happen.

Please let me know if you need any more information.
Cheers,
Josh

sungam3r:
@JoshKeegan Please rebase onto `master` if you are going to continue work on this PR.

JoshKeegan:
I'm not using this functionality any more, and @unaizorrilla's last comment questions whether the proposed behaviour is correct.
So if nothing has changed with the thinking around exceptions during wireup, then I'm happy for this to be closed/abandoned.

sungam3r:
OK

Issue #273, opened 2019-09-19
HealthChecks.EventStore under Linux: connection leak


Zetanova:
HealthChecks.EventStore lets the connection open or does not properly dispose it. 
There is a know issue with sockets under linux, 
where undisposed sockets don't get finalized by the GC and stack up over time. 

HttpClient's and other components fill then fail with "address already in use"

unaizorrilla:
Follow this build https://ci.appveyor.com/project/Xabaril/aspnetcore-diagnostics-healthchecks/builds/27602790 to see when the package will be created, new version for EventStore is 2.2.3

PR #277, opened 2019-09-23
EventStore does not check the connection itself

Zetanova:
The eventstore client method ```await connection.ConnectAsync()``` already completes 
after the init of the TcpConnection instance. It is then in the internal State "Connecting" 
and it is not indicating a success or failure.

To check the real result of the connection,
an event handler to the event Connected, Closed and ErrorOccurred
is required. 

Taken from the UnitTest of Eventstore:
https://github.com/EventStore/EventStore/blob/master/src/EventStore.Core.Tests/ClientAPI/connect.cs 

Maybe it will fix related issue
https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/273


unaizorrilla:
Merged with minor changes, thanks for contribute @Zetanova

Issue #237, opened 2019-07-19
Enable Health Check Release Gate to execute as a task


carlowahlstedt:
I setup a health check release gate to execute as a post deployment gate and it started failing. I eventually realized that the url I was trying to hit isn't public and therefore because the gate doesn't run on my on-prem agent that it's not going to succeed. I'd like to suggest enabling the task to be able to run as a task inside of a pipeline. If you agree this is a good idea, I'm willing to look at what we need to do. I've got some experience building tasks.

CarlosLanderas:
Ping @lfraile

lfraile:
Sorry I didn't read this before. 

Seems reasonable. All the code is in the repo so if you want to take a look go ahead. Let me know if you need help. We can do a remote coding session if needed.

sungam3r:
https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/291#issuecomment-1050354157

PR #291, opened 2019-10-02
initial go at adding it as a task

carlowahlstedt:
First attempt at #237. I'm not familiar with how to use JSONPath and got stuck, so decided it best to get a first review for continuing on. Let me know if you agree with the direction here and any ideas/pointers you have.


unaizorrilla:
Hi @lfraile  can you take a look on this PR?

sungam3r:
@carlowahlstedt Is it still actual?

carlowahlstedt:
Going to be honest. I'm so far away from this at this point that I barely remember what this was about.

sungam3r:
OK. That happens.

PR #301, opened 2019-10-13
Removing exception handling

vitskr:
Exception returned by CheckHealthAsync in HealthCheckResult is not logged by DefaultHealthCheckService. It is perfectly fine for CheckHealthAsync to throw exception. It results in HealthCheckResult.Unhealthy and exception is logged too
 
For more information check:
https://github.com/aspnet/Extensions/blob/1dfe8bc2af511822677789c29a0c800f7cca1d7b/src/HealthChecks/HealthChecks/src/DefaultHealthCheckService.cs#L137

### DIFF
--- a/src/HealthChecks.SqlServer/SqlServerHealthCheck.cs
+++ b/src/HealthChecks.SqlServer/SqlServerHealthCheck.cs
@@ -18,24 +18,17 @@ namespace HealthChecks.SqlServer
         }
         public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
         {
-            try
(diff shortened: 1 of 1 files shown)


unaizorrilla:
Hi @vitskr 

I know the code inside DefaultHealthCheckService! the reason for this catch is honoring the Registration.FailureStatus, and is the recomended way!

I close the issue, but you can comment if need more conversation about this!

vitskr:
That's unfortunate. No way to log exception thrown inside SqlServerHealthCheck.CheckHealthAsync

unaizorrilla:
You can add ILogger and log the exception, we accept a PR about this!

Issue #299, opened 2019-10-09
AppInsights Publisher doesn't track exceptions for health results with exceptions


inkel:
I've identified that the current `ApplicationInsightsPublisher` https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/b31936bed18f33f8f3c1dd65180911b6affbd92a/src/HealthChecks.Publisher.ApplicationInsights/ApplicationInsightsPublisher.cs do not track exceptions when a report contains exceptions.

Most, if not all, of the implemented health checks, have this form:

```csharp
        public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
        {
            try
            {
                /* health check logic here */
            catch (Exception ex)
            {
                return new HealthCheckResult(context.Registration.FailureStatus, exception: ex);
            }
        }
```

But the publisher is ignoring the value of `HealthCheckResult.Exception` when tracking events.

I already have a very crude local branch that adds an additional `bool trackExceptions = true`  parameter to the `AddApplicationInsightsPublisher` extension method. Are you interested in a PR to add this feature?

inkel:
I'm closing this one as #302 has been merged.

sibbl:
A small follow-up question: has this already been released half a year later? I don't see the `trackExceptions` parameter in version 3.1.1.

Please find a very small ASP.NET Core repro project here:
[ReproHealthChecksAppInsightsPublisher.zip](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/files/4597619/ReproHealthChecksAppInsightsPublisher.zip)

I hope it's okay that I didn't create a new issue.

inkel:
Came here after someone left a comment (now deleted):

> A small follow-up question: has this already been released half a year later? I don't see the trackExceptions parameter in version 3.1.1.

This got me curious so I when to check, and it seems that the code that was merged in #302 is not present anymore. I did some investigation and it seems that the exception tracking code was removed in commit 8748404d925d38edb9f051733bbb7439eaaeed03. Sorry if I'm stepping out of line, but can you explain why it was removed? It seems like a regression that the code is back to not reporting exceptions to Application Insights.

inkel:
Any updates on this one?

PR #302, opened 2019-10-15
Track exceptions in Application Insights publisher

inkel:
This PR implements exception tracking on App Insights as described in #299. 

My initial reaction was to add this only in the detailed report, but then I realized that I wanted to have the exceptions tracked, regardless of whether it's a detailed or summarized report.

While working on this I also found that it could be possible to refactor the code to be more DRY (e.g. creating the properties dictionary could be extracted to a method), and also that by using a mock `TelemetryChannel` tests could be added for this publisher. I didn't include those changes because I think they're out of the scope of this PR description.

### DIFF
--- a/src/HealthChecks.Publisher.ApplicationInsights/ApplicationInsightsPublisher.cs
+++ b/src/HealthChecks.Publisher.ApplicationInsights/ApplicationInsightsPublisher.cs
@@ -23,8 +23,9 @@ namespace HealthChecks.Publisher.ApplicationInsights
         private static readonly object sync_root = new object();
         private readonly bool _saveDetailedReport;
         private readonly bool _excludeHealthyReports;
+        private readonly bool _trackExceptions;
 
-        public ApplicationInsightsPublisher(string instrumentationKey = default, bool saveDetailedReport = false, bool excludeHealthyReports = false)
+        public ApplicationInsightsPublisher(string instrumentationKey = default, bool saveDetailedReport = false, bool excludeHealthyReports = false, bool trackExceptions = true)
         {
             _instrumentationKey = instrumentationKey;
             _saveDetailedReport = saveDetailedReport;
@@ -48,7 +49,30 @@ namespace HealthChecks.Publisher.ApplicationInsights
                 SaveGeneralizedReport(report, client);
             }
 
+            if (_trackExceptions)
+            {
+                TrackExceptions(report, client);
+            }
+
             return Task.CompletedTask;
+        }
+        private void TrackExceptions(HealthReport report, TelemetryClient client)
+        {
+            foreach (var reportEntry in report.Entries.Where(entry => entry.Value.Exception != null))
+            {
+                client.TrackException(reportEntry.Value.Exception,
+                    properties: new Dictionary<string, string>()
+                    {
+                        { nameof(Environment.MachineName), Environment.MachineName },
+                        { nameof(Assembly), Assembly.GetEntryAssembly().GetName().Name },
+                        { HEALTHCHECK_NAME, reportEntry.Key }
+                    },
+                    metrics: new Dictionary<string, double>()
+                    {
+                        { METRIC_STATUS_NAME, reportEntry.Value.Status == HealthStatus.Healthy ? 1 : 0 },
+                        { METRIC_DURATION_NAME, reportEntry.Value.Duration.TotalMilliseconds }
+                    });
+            }
         }
         private void SaveDetailedReport(HealthReport report, TelemetryClient client)
         {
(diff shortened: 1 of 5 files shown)


src/HealthChecks.Publisher.ApplicationInsights/ApplicationInsightsPublisher.cs
```diff
@@ -23,8 +23,9 @@ class ApplicationInsightsPublisher
         private static readonly object sync_root = new object();
         private readonly bool _saveDetailedReport;
         private readonly bool _excludeHealthyReports;
+        private readonly bool _trackExceptions;
 
-        public ApplicationInsightsPublisher(string instrumentationKey = default, bool saveDetailedReport = false, bool excludeHealthyReports = false)
+        public ApplicationInsightsPublisher(string instrumentationKey = default, bool saveDetailedReport = false, bool excludeHealthyReports = false, bool trackExceptions = true)
```
unaizorrilla: Where is _tracekExpcetions assigned?

src/HealthChecks.Publisher.ApplicationInsights/ApplicationInsightsPublisher.cs
```diff
@@ -48,8 +49,31 @@ public Task PublishAsync(HealthReport report, CancellationToken cancellationToke
...
             }
 
+            if (_trackExceptions)
+            {
+                TrackExceptions(report, client);
+            }
+
             return Task.CompletedTask;
         }
+        private void TrackExceptions(HealthReport report, TelemetryClient client)
```
unaizorrilla: If the health report contains exception, you can save also detailed report and exception? probably detailed and generalized report will contain exception information without create any other method

src/HealthChecks.Publisher.ApplicationInsights/ApplicationInsightsPublisher.cs
unaizorrilla: How you can change TrackException parameters? probably you need to add a new overload or default  parameter on  ApplicationInsightsHealthCheckBuilderExtensions.cs

src/HealthChecks.Publisher.ApplicationInsights/ApplicationInsightsPublisher.cs
inkel: That's a good question. Initially, I was going to `TelemetryClient.TrackException` only in the detailed report, but I think it's information you would like to have even in the summary report, that's why I've added the `TrackExceptions` method.

Also, note that for detailed report you could choose to not report healthy results; it probably doesn't make much sense but it could happen that a report is healthy but it also includes an exception and this is why the `TrackExceptions` method is called regardless you choose a summary or detailed report.

src/HealthChecks.Publisher.ApplicationInsights/ApplicationInsightsPublisher.cs
inkel: You are correct, I was going to add a default on `ApplicationInsightsHealthCheckBuilderExtensions`. Again, the plan is adding a default `bool trackExceptions = true` rather than false because I think having a report on exceptions happened is something one would like for default (at least I know it's what I wanted).

inkel:
@unaizorrilla thanks for the comments! I've left some answers, and I'll be working on some of those suggestions.

I'll also try to add a basic test for this new operation.

src/HealthChecks.Publisher.ApplicationInsights/ApplicationInsightsPublisher.cs
inkel: This has been fixed in 4d93ae545bcaac05e3b7c4f2fd6e0ed686993f73.

inkel: I've add the parameter to the extension method in 0c2e3894765a1c0bbcef39fb7f6da4c996f00410.

inkel:
@unaizorrilla I've made some changes and added some tests, I think it's ready for another round of review. Thank you!

unaizorrilla:
Hi @inkel 

I try to review this again today!

unaizorrilla:
I'm so sorry for delaying on review the PR! :-(  I try to do ASAP

unaizorrilla:
Hi @inkel 

I perform some changes in the pr and merge it to master, many thanks to contribute on this.

My changes try to preserve your intention, track exception on ai also as part of telemetry information. But I consider to do only when you set detailed reports and not a separate method.

Also I need to work more on unit test to improve our coverage...

Thanks again to contribute and so sorry to do changes in your pr withut normal way

inkel:
@unaizorrilla hi, thanks! I'm more than ok with you making changes 😸 though for some reason GitHub is not showing those to me. Anyway, thanks for being open to receive, review and merge this PR ❤️ 

> Also I need to work more on unit test to improve our coverage...

Would you like me to help you with that? One of the reasons I've added the tests for this publishe was that secretly I planned on submitting another PR that added more tests to it for the existing features 😊

Issue #304, opened 2019-10-16
Re-use rabbit connection for RabbitMq Health check


jnovick:
As per [official docs](https://www.rabbitmq.com/connections.html) and its suggestions on [high connectivity churn](https://www.rabbitmq.com/networking.html#dealing-with-high-connection-churn), connections are meant to be long lived. Ideally, I believe this should be configured as a singleton, but I understand there may be use cases where this could be configured differently. Currently the code does a [new connection per health check request](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/706fb5b7c1991272ab73aa00a3e868c03589d93f/src/HealthChecks.RabbitMQ/RabbitMQHealthCheck.cs#L28). I believe this should use an existing connection if there is already a connection registered via DI or create one to be used for all health check requests instead of creating a new one each time.

Edit: This is mainly an issue when running in an environment like Kubernetes that regularly hits the health check endpoint. In Kubernetes, the default is 10s as per [k8s official docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes). My readiness probe hits an endpoint that uses this health check and that is why it is creating such an issue for me.

unaizorrilla:
Hi @jnovick 

We do some changes in other healtchecks related with the same issue, we are happy to accept a PR about his if you have some time, if not, I can do this but need some time!

jnovick:
Okay, I'll submit a PR shortly. I am planning on adding several options to allow users to solve the problem in the best way that fits their project. I don't want to lock people into fitting their project architecture to fit the same way I have mine by designing the health check to solve my problem. Let me know what you think once I push up a PR.

jnovick:
I have submitted a PR (https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/307). Let me know what you think. I prioritized not changing existing behavior and instead making it easy to follow a better pattern, but I am not sure if I created too many config options.

rwkarg:
The Func based extension methods introduced in #307 don't actually result in connection reuse. A new RabbitMQHealthCheck instance is created from the factory each time the HealthCheck infrastructure evaluates.

PR #352 fixes this for the connection string based extension method, but the `Func<IServiceProvider, >` based extension methods will always result in the provided factory being called each time a health check is evaluated and the implementation of those extension methods always creates a new RabbitMQHealthCheck instance.
[...]

PR #307, opened 2019-10-20
Add ability to re-use single RabbitMQ Connection

jnovick:
* Addresses https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/304
* Add many configuration options that support ability to re-use a single
connection with multiple multiplexed channels.
* Add documentation around this health check to help others follow
RabbitMQ best practices

### DIFF
--- a/src/HealthChecks.RabbitMQ/DependencyInjection/RabbitMQHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.RabbitMQ/DependencyInjection/RabbitMQHealthCheckBuilderExtensions.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Extensions.DependencyInjection
         const string NAME = "rabbitmq";
 
         /// <summary>
-        /// Add a health check for RabbitMQ services.
+        /// Add a health check for RabbitMQ services using connection string (amqp uri).
         /// </summary>
         /// <param name="builder">The <see cref="IHealthChecksBuilder"/>.</param>
         /// <param name="rabbitMQConnectionString">The RabbitMQ connection string to be used.</param>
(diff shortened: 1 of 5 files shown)


src/HealthChecks.RabbitMQ/DependencyInjection/RabbitMQHealthCheckBuilderExtensions.cs
```diff
@@ -33,5 +33,92 @@ public static IHealthChecksBuilder AddRabbitMQ(this IHealthChecksBuilder builder
...
+        /// <param name="builder">The <see cref="IHealthChecksBuilder"/>.</param>
+        /// <param name="name">The health check name. Optional. If <c>null</c> the type name 'rabbitmq' will be used for the name.</param>
+        /// <param name="failureStatus">
+        /// The <see cref="HealthStatus"/> that should be reported when the health check fails. Optional. If <c>null</c> then
+        /// the default status of <see cref="HealthStatus.Unhealthy"/> will be reported.
+        /// </param>
+        /// <param name="tags">A list of tags that can be used to filter sets of health checks. Optional.</param>
+        /// <param name="timeout">An optional System.TimeSpan representing the timeout of the check.</param>
+        /// <returns>The <see cref="IHealthChecksBuilder"/>.</returns></param>
+        public static IHealthChecksBuilder AddRabbitMQ(this IHealthChecksBuilder builder, string name = default,
```
jnovick: One problem here is that there is ambiguity between the existing method's signature and this one if you are only specifying a single string. I can either rename this method or consolidate into a single method where it has different behavior depending on what fields are provided to prevent this ambiguity.

src/HealthChecks.RabbitMQ/DependencyInjection/RabbitMQHealthCheckBuilderExtensions.cs
```diff
@@ -33,5 +33,92 @@ public static IHealthChecksBuilder AddRabbitMQ(this IHealthChecksBuilder builder
...
+            return builder.Add(new HealthCheckRegistration(
+                name ?? NAME,
+                sp => {
+                    var connection = sp.GetService<IConnection>();
+                    var connectionFactory = sp.GetService<IConnectionFactory>();
+                    if (connection != null)
+                    {
+                        return new RabbitMQHealthCheck(connection);
+                    }
+                    else if(connectionFactory != null)
```
sungam3r: whitespace

src/HealthChecks.RabbitMQ/README.md
```diff
@@ -0,0 +1,130 @@
...
+        AutomaticRecoveryEnabled = true
+    };
+
+    services
+        .AddHealthChecks()
+        .AddRabbitMQ(sp => factory);
+}
+```
+
+### Without Conveniece Extension Method
```
sungam3r: Convenience

src/HealthChecks.RabbitMQ/README.md
```diff
@@ -0,0 +1,130 @@
...
+```cs
+public void ConfigureServices(IServiceCollection services)
+{
+    var factory = new ConnectionFactory()
+    {
+        Uri = new Uri("amqps://user:pass@host/vhost"),
+        AutomaticRecoveryEnabled = true
+    };
+
+    var connection = factory.CreateConnection();
```
ryepup: I know this is only some sample code, but people frequently copy/paste these things and assume the code fit for production. In this case, I don't think we want to be creating a real connection in `ConfigureServices`. We want our health check to fail, and as written I think the app will fail to start.

I think we can keep this connection lazy without too much hassle using another `AddSingleton` overload:

```csharp
.AddSingleton<IConnection>(_ => factory.CreateConnection())
```

unaizorrilla:
Hi @jnovick 

I will try to review this PR today afternoon!

src/HealthChecks.RabbitMQ/DependencyInjection/RabbitMQHealthCheckBuilderExtensions.cs
ryepup: We could resolve the ambiguity by re-ordering the parameters here, but that feels a little cheaty:

```csharp
public static IHealthChecksBuilder AddRabbitMQ(this IHealthChecksBuilder builder, HealthStatus? failureStatus = default, string name = default, 
```

I think with that ordering a call like `AddRabbitMQ("amqps://broker")` would resolve to the overload on L27, and if we wanted to call this overload we'd need something like `AddRabbitMQ(name: "broker")`

jnovick: The thought I had today was to add a bool indicating whether to give priority to IConnectionFactory instead of IConnection. If I make it optional. I introduce a possibly useful feature and I maintain backwards compatibility.

ryepup: that'd do it; what would that parameter be called? `preferIConnectionFactoryFromServices` ?

the code might get a little messy to make this work as a priority; how about making it a little narrower, something like: `bool useIConnectionFactoryFromServices=false`. If it's true, we only look for an `IConnectionFactory`, if it's false, we only look for an `IConnection`?

unaizorrilla:
Hi @jnovick 
Merged on master, many thanks for improve a lot this health check ( top downloaded )! I add some missing tests for cover different registration options and solve a minor issue.

After CI build I will create a new tag and when the release build finish a new package will be created with version (3.0.1)

P.S: I change also version for us 3.0 is the .NET Core Version, and  next is our changes, is not semver!

unaizorrilla:
Release build finished, 3.0.1 package for Rabbitmq is on Nuget

PR #328, opened 2019-11-12
Dynamodb credentials

hpcsc:
This PR does a few things:
- Allow DynamoDB healthcheck to take in `ServiceUrl` in addition to `RegionEndpoint`. This is so that DynamoDb Healthcheck can be run against local DynamoDB
- Instead of always using `AccessKey` and `SecretKey`, provide an option to let AWS SDK looks up credentials from its credentials provider chain. This is much more flexible for different scenarios (.e.g. when running in AWS environment and credentials are from role/instance profile, not explicitly provided)
- Add unit and functional tests


hpcsc:
is there anyway to troubleshoot the build in appveyor? The tests are failing there but it's working fine for me in local

hpcsc:
look like when building in appveyor, it doesn't start dependencies in `docker-compose`: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/b13326f641aed54c9e74c5e43de5408602ad5cca/build.ps1#L52

then the tests are failing are due to new local DynamoDB container in `docker-compose` is not started in appveyor. Does anyone know how to add a container dependency for functional tests in appveyor?

PR #329, opened 2019-11-13
Add support for Azure Service Bus queues with Session enabled

rnarayana:
Hi,
When the Azure Service Bus has sessions enabled, it does not allow messages to be added without setting the SessionId, and results in an Exception. I've made a small modification to the Service Bus queue health check to pass a SessionId string when needed

Please review and let me know!

Thanks,
Narayana.


src/HealthChecks.AzureServiceBus/AzureServiceBusQueueHealthCheck.cs
```diff
@@ -45,9 +48,17 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
                     {
                         return new HealthCheckResult(context.Registration.FailureStatus, description: "New QueueClient connection can't be added into dictionary.");
                     }
+
+                    var managementClient = new ManagementClient(_connectionString);
+                    var queueDescription = await managementClient.GetQueueAsync(_queueName);
```
sungam3r: What happens if the queue `_queueName` does not exist?

sungam3r: MessagingEntityNotFoundException("Queue was not found");

rnarayana: Yes, it is also the same exception that will be thrown on a call to queueClient.ScheduleMessageAsync() passing a non-existent queue (Surprisingly, you can create a QueueClient passing in a non-existent queue).
Should I handle a case where there is no queue during the first call, but gets created eventually? Having _requiresSession as static would pose problems in that case.

rnarayana: I've updated to handle cases where queue is deleted or created after health check gets run once. This would be useful if this is part of a readiness probe in kubernetes, where the container is not killed on health check failure.

sungam3r: > Should I handle a case where there is no queue during the first call, but gets created eventually?

Сontrol will fall into catch block so I think there is no difference.

src/HealthChecks.AzureServiceBus/AzureServiceBusQueueHealthCheck.cs
```diff
@@ -12,10 +13,17 @@ public class AzureServiceBusQueueHealthCheck
...
         private static readonly ConcurrentDictionary<string, QueueClient> _queueClientConnections = new ConcurrentDictionary<string, QueueClient>();
 
         private readonly string _connectionString;
         private readonly string _queueName;
+
+        /// <summary>
+        /// Null value for this field specifies not being able to connect to queue yet to know session requirement.
+        /// This could happen if queue is not yet created or gets deleted, after health check starts.
+        /// </summary>
+        private bool? _requiresSession;
```
unaizorrilla: private members does not need summary comments

unaizorrilla: who is set _requiresSession? probably you need to add new parameter into AddXX to include this new option?

src/HealthChecks.AzureServiceBus/AzureServiceBusQueueHealthCheck.cs
```diff
@@ -39,25 +47,40 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
...
                         return new HealthCheckResult(context.Registration.FailureStatus, description: "New QueueClient connection can't be added into dictionary.");
                     }
                 }
 
+                // If the queue is not yet available, GetQueueAsync will throw and _requiresSession will not have any value
+                if (!_requiresSession.HasValue)
+                {
+                    var managementClient = new ManagementClient(_connectionString);
+                    var queueDescription = await managementClient.GetQueueAsync(_queueName);
+                    _requiresSession = queueDescription.RequiresSession;
```
unaizorrilla: Well, the problem here is you use management to get information about if the queue require session, we prefer add new flag into AddXXX because query management is overhead  in this process

rnarayana: My thought was this will be done only the first time, and hence the overhead should be fine - once the requiresSession is set, ManagementClient will not be used. The advantage is that the user does not have to explicitly pass a requiresSession, and can be figured out from the queue. However if you prefer passing it explicitly, I can either create an optional session argument defaulted to false in ctor, or a new AddSession() method.

unaizorrilla: Sounds good option session defaulted to false, or some like Options to configure this like this

unaizorrilla: Please tagme when push new commits on this to review and merge

rnarayana: @unaizorrilla Can you please review? Thanks!

CarlosLanderas:
@rnarayana  you exposed requiresSession parameter in HealthCheck constructor but it's not exposed in AddAzureServiceBusQueue extension method so it's always going to be false. Could you fix that?

rnarayana:
> 
> 
> @rnarayana you exposed requiresSession parameter in HealthCheck constructor but it's not exposed in AddAzureServiceBusQueue extension method so it's always going to be false. Could you fix that?

Done. (Please do squash the commits if you are merging)

CarlosLanderas:
Tested and working. Thanks for the contribution @rnarayana

PR #332, opened 2019-11-13
[Docker Image] Custom branding stylesheet and Paths environment variables

CarlosLanderas:
With version 3.0.3 the UI can be configured to use a custom css stylesheet for branding using a volume and an environment variable and it can also receive different env vars to configure ui, api an resources paths

### DIFF
--- a/build/docker-images/HealthChecks.UI.Image/Extensions/IEndpointRouteBuilderExtensions.cs
+++ b/build/docker-images/HealthChecks.UI.Image/Extensions/IEndpointRouteBuilderExtensions.cs
@@ -1,0 +1,24 @@
+﻿using HealthChecks.UI.Image.Configuration;
+using HealthChecks.UI.Image.Extensions;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.Configuration;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static  class IEndpointRouteBuilderExtensions
+    {
+        public static IEndpointConventionBuilder MapDockerHealthchecksUI(this IEndpointRouteBuilder builder, IConfiguration configuration)
+       {
+            return builder.MapHealthChecksUI(setup =>
+            {
+                setup.ConfigureStylesheet(configuration);
+                setup.ConfigurePaths(configuration);
+            });
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


build/docker-images/HealthChecks.UI.Image/Extensions/IEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.Configuration;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static  class IEndpointRouteBuilderExtensions
```
sungam3r:
```suggestion
    public static class IEndpointRouteBuilderExtensions
```

CarlosLanderas: Thanks @sungam3r

build/docker-images/HealthChecks.UI.Image/Extensions/IEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public static IEndpointConventionBuilder MapDockerHealthchecksUI(this IEndpointRouteBuilder builder, IConfiguration configuration)
```
sungam3r:
```suggestion
        public static IEndpointConventionBuilder MapHealthchecksUI(this IEndpointRouteBuilder builder, IConfiguration configuration)
```

sungam3r: In fact there is nothing special about Docker, just "standard" configuring.

CarlosLanderas: That method already exists. The intention is create a method to configure docker related stuff that can not be configured as the image is already biult

sungam3r: > That method already exists in UI assembly with Microsoft.AspNetCore.Builder namespace

With different signature. It will be just another overload.

> to configure docker related stuff 

This is clear. But the code itself has nothing to do with Docker. This is the usual `IConfiguration`-based configuration. Such a name can confuse someone.

CarlosLanderas: Ok @sungam3r, it makes sense. updated

Issue #279, opened 2019-09-24
AddHealthCheckEndpoint  Question


eobeda:
Feature: HealthChecks-UI

When using the AddHealthCheckEndpoint setting, I'd like to use a relative path or find a way to extract the IIS in-process address and port that the site in running under.
```
 services.AddHealthChecksUI("healthchecksdb", setupSettings =>
 {
     setupSettings.AddHealthCheckEndpoint("Name", "url relative path");
 }
```
I found the options for "UseRelativeApiPath" during config, but setting it to true doesn't seem to control the AddHealthCheckEndpoint  parsing
```
  config.MapHealthChecksUI(setup =>
                    {          
                        setup.UseRelativeApiPath=true;
```

Any suggestions ?
Thanks
Ed
ps: Using VS 2019 16.3 with Asp.Net Core 3.0 and custom branding of the HealthChecks-UI web page. AspNetCore.HealthChecks.UI  is version 3.0

unaizorrilla:
Hi @eobeda 

thanks for ask the question, @CarlosLanderas can you take a look here!

CarlosLanderas:
@eobeda , I know you can get the listening urls using 
```csharp
app.ServerFeatures.Get<IServerAddressesFeature>().Addresses;
```
You can get several addresses (http, https) and maybe use the first to get a relative url 
but this does not work with IIS, is a feature that only reports the listening addresses when using Kestrel.

We would need to have this feature working along with kestrel and IIS

Maybe @rynowak  can give us some more information about this, I don't know if this is possible from code when using IIS

CarlosLanderas:
@rynowak @glennc any Idea about how this can be achieves when using IIS?

rynowak:
@tratcher @jkotalik - any ideas?

Tratcher:
IServerAddresses was added to IIS in-proc in 3.0. That said, using a relative path is a much better idea. What do the expected urls look like for these various components? Generating a url per-request is much easier because you can use the request's scheme, host, port, PathBase, etc..

CarlosLanderas:
@rynowak @Tratcher thank you very much for your help :)

CarlosLanderas:
@Tratcher is this working with IIS Express?. I'm doing a test and I get 0 ServerAddresses configuring IIS Express InProcess

![image](https://user-images.githubusercontent.com/16475649/68924441-215add00-0781-11ea-9e81-a8689f194dd1.png)

[...]

Issue #38, opened 2018-12-17
Allow to use relative address in Uri


cilerler:
As you may see below, serving address may change depends on environment. 
Unfortunately adding multiple references and/or changing it constantly is not a pleased workaround.

It would be convenient to have an option to use relative address in `Uri` like we can in **launchSettings.json**'s `launchUrl` property.

## appsettings.json

```json
"HealthChecks-UI": {
	"HealthChecks": [
		{
			"Name": "Self",
			"Uri": "http://localhost:47920/health/beatpulse"
		}
	]
}
```

## launchSettings.json
```
{
	"iisSettings": {
		"windowsAuthentication": false,
		"anonymousAuthentication": true,
		"iisExpress": {
			"applicationUrl": "http://localhost:47920/",
			"sslPort": 0
		}
	},
	"$schema": "http://json.schemastore.org/launchsettings.json",
	"profiles": {
		"IIS Express": {
			"commandName": "IISExpress",
			"launchBrowser": true,
			"launchUrl": "health/beatpulse-ui",
			"environmentVariables": {
				"ASPNETCORE_ENVIRONMENT": "Development"
			}
		},
		"MyProject.Api": {
			"commandName": "Project",
			"launchBrowser": true,
			"launchUrl": "health/beatpulse-ui",
			"environmentVariables": {
				"ASPNETCORE_ENVIRONMENT": "Development"
			},
			"applicationUrl": "https://localhost:47923/;http://localhost:47922/"
		},
		"Docker": {
			"commandName": "Docker",
			"launchBrowser": true,
			"launchUrl": "{Scheme}://{ServiceHost}:{ServicePort}/api/values"
		  }
	}
}
```

CarlosLanderas:
@cilerler this scenario might be only supported when you are configuring the UI on the same place as the HealthChecks. It that you use case? For separate hosts the relative Uri would not be valid.

cilerler:
correct

CarlosLanderas:
@cilerler I was working on other feature but will try to implement this soon :)

CarlosLanderas:
@cilerler  using the server feature `IServerAddressesFeature` we can get the configured urls when using **Kestrel** without problems to allow relative urls in settings but I am not really sure if we can do that when using **IIS**.

Like this is happening on startup and we do not have an active request we can just relay on information provided by the server and when using this feature IIS returns null.

@unaizorrilla  any idea?.

[...]

PR #339, opened 2019-11-16
[UI] Allow relative urls in healthchecks and webhooks endpoints using IServerAddressesFeature

CarlosLanderas:
**Motivation**

Some people have requested the possibility to add relative urls in healthchecks and webhooks configuration, by using setup actions or appsettings.json. 

#38 
#279 

Some users are setting the UI in the same endpoint as the healthchecks and when they are deploying to the cloud they don't even know the port the server will use or they just don't wont to worry about configuring absolute urls with ports thats might change in the future.

```csharp

setup.AddHealthCheckEndpoint("endpoint2", "/healthz");
setup.AddWebhookNotification("webhook1", uri:  "/status/200?code=ax3rt56s", payload: ...);
```

**Idea**

The idea is detecting the urls the server is listening on and compose the absolute urls using the urls reported by the webserver to be used by the HealthCheckReportCollector.

Until .net core 3.0, we had a problem with IIS, because the server feature called IServerAddressesFeature was only working with Kestrel but with latest core version, if you use hosting model InProcess  (Default) you will have this addresses reported as well and thats why this can be achieved right now.

**Problems**

- The main problem I found is when using IIS, this addresses are reported after the webserver has started. It has some delay, so if the HealthCheckReportCollectorHostedService starts and the collector tries to query the remote endpoints too soon with relative urls, and the listening addresses are not yet reported this causes a race condicition and we can't parse the relative to absolute urls.

To solve this problem where failed checks might appear in the UI due to wrong formed urls, I decided the hostedservice not to trigger collections until the ServerAddressService knows this listening addreses and I want some kind of opinion because I don't know if this is the best way to achieve it.

- If the user configures IIS hosting model to Out of Process, the urls reported by the IServerAddressesFeature are the kestrel reverse proxies ones and the final urls will not work.
Default hosting model for IIS is In Process, so all users not changing hosting model to Out Of Process will have this feature working. One possibility would be explaining in documentation that the UI will only work with In Process hosting model or adding a new commit to explicitely enable/disable relative uris support.

**Testing desired!**

As this affects how the UI internally triggers healthcheck collection, it would be great to be tested by several people using different configurations (Kestrel, IIS In Process) to see that everything is working ok. In my case, the feature is working without problems except for the Out Of process IIS hosting model


src/HealthChecks.UI/Core/HostedService/HealthCheckReportCollectorHostedService.cs
```diff
@@ -49,20 +55,24 @@ private async Task ExecuteAsync(CancellationToken cancellationToken)
 
             while (!cancellationToken.IsCancellationRequested)
             {
-                _logger.LogDebug("Executing HealthCheck collector HostedService.");
-
-                using (var scope = scopeFactory.CreateScope())
+                // Collect should not be triggered until IServerAddressFeature reports the listening endpoints
```
CarlosLanderas: If we do not want a race condition where the hosted service trigger collections using relative urls and the webserver has not reported the listening urls yet we should control that the ServerAddressesService already know this urls

src/HealthChecks.UI/Core/HostedService/HealthCheckReportCollector.cs
```diff
@@ -89,6 +96,26 @@ private async Task<UIHealthReport> GetHealthReport(HealthCheckConfiguration conf
                 return UIHealthReport.CreateFrom(exception);
             }
         }
+
+        private Uri GetEndpointUri(HealthCheckConfiguration configuration)
+        {
+             if (endpointAddresses.ContainsKey(configuration.Id))
```
CarlosLanderas: As healthchecks trigger far more often than webhooks, we hold parsed absolute uris in an in memory dictionary to avoid parsing of all the endpoints that have already been triggered

test/FunctionalTests/HealthChecks.UI/ServerAddressesServiceTests.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            serverAddressService.AbsoluteUriFromRelative("/my/relative/url")
+               .Should().Be($"{serverAddress}/my/relative/url");
+
+            serverAddressService.AbsoluteUriFromRelative("segment1/segment2/segment3")
+             .Should().Be($"{serverAddress}/segment1/segment2/segment3");
+        }
+    }
+
+    internal class CustomServerAddressFeature : IServerAddressesFeature
```
CarlosLanderas: TestServer does not register Feature Collections so we simulate a feature that will report the urls the same way as IIS and Kestres do.

src/HealthChecks.UI/Configuration/Settings.cs
```diff
@@ -27,12 +27,10 @@ public Settings AddHealthCheckEndpoint(string name, string uri)
         
         public Settings AddWebhookNotification(string name, string uri, string payload, string restorePayload = "")
         {
-            if (!Uri.TryCreate(uri, UriKind.Absolute, out Uri absoluteUri)) throw new ArgumentException($"Invalid uri: {uri}");
```
CarlosLanderas: This check is removed as Uris processing will come later when parsing relative / absolute urls

CarlosLanderas: @Tratcher @jkotalik @rynowak this is the part of the PR I am most concerned about as I do not know if they is another way of waiting until the listening urls are reported by the IServerAddressesFeature when using IIS with InProcess hosting model. 

Without this knowledge, I can just delay the healthcheck collection execution until the ServerAddressesService knows this listerning addresses.

Could you give some advice?

src/HealthChecks.UI/Core/HostedService/HealthCheckReportCollectorHostedService.cs
CarlosLanderas: @Tratcher @jkotalik @rynowak this is the part of the PR I am most concerned about as I do not know if they is another way of waiting until the listening urls are reported by the IServerAddressesFeature when using IIS with InProcess hosting model. 

Without this knowledge, I can just delay the healthcheck collection execution until the ServerAddressesService knows this listerning addresses.

Could you give some advice?

src/HealthChecks.UI/Core/HostedService/HealthCheckReportCollectorHostedService.cs
```diff
@@ -49,20 +55,24 @@ private async Task ExecuteAsync(CancellationToken cancellationToken)
 
             while (!cancellationToken.IsCancellationRequested)
             {
-                _logger.LogDebug("Executing HealthCheck collector HostedService.");
-
-                using (var scope = scopeFactory.CreateScope())
+                // Collect should not be triggered until IServerAddressFeature reports the listening endpoints
+                if (_serverAddressesService.HasAddresses)
```
CarlosLanderas: Do not trigger any endpoint collection until the addresses service knows the urls the server is listening on.

src/HealthChecks.UI/Core/HostedService/HealthCheckReportCollectorHostedService.cs
```diff
@@ -49,20 +55,24 @@ private async Task ExecuteAsync(CancellationToken cancellationToken)
...
+                        try
+                        {
+                            var runner = scope.ServiceProvider.GetRequiredService<IHealthCheckReportCollector>();
+                            await runner.Collect(cancellationToken);
+
+                            _logger.LogDebug("HealthCheck collector HostedService executed successfully.");
+                        }
+                        catch (Exception ex)
+                        {
+                            _logger.LogError(ex, $"HealthCheck collector HostedService throw a error: {ex.Message}");
```
rynowak: BTW strings used with logging shouldn't use `$""`. Logging has its own format for capturing values.

```C#
_logger.LogError(ex, $"HealthCheck collector HostedService throw a error: {ExceptionMessage}, ex.Message, ex);
```

src/HealthChecks.UI/Core/HostedService/HealthCheckReportCollectorHostedService.cs
rynowak: @davidfowl helped figure out a workaround.

```C#
    public class AddressHostedService : IHostedService
    {
        private readonly IServer _server;
        private readonly IHostApplicationLifetime _lifetime;

        public AddressHostedService(IServer server, IHostApplicationLifetime lifetime)
        {
            _server = server;
            _lifetime = lifetime;
            _lifetime.ApplicationStarted.Register(() =>
            {
                var addresses = _server.Features.Get<IServerAddressesFeature>().Addresses.ToArray();
            });
        }
    }
```

The `ApplicationStarted` cancellation token will trigger once all of the hosted services have initialized. This way you can be notified once everything is *fully* started.

----

The problem in your case is that there's an order dependency between your hosted service and the "web host" hosted service. By default if you register a hosted service from `ConfigureServices` it's going to initialize before the web host does.

CarlosLanderas: @rynowak and @davidfowl thank you so much for your help!. I'm going to try it this way 👍

CarlosLanderas: @rynowak @Davidfowl, I still have the same race condition using ApplicationStarted because as you explained,  ApplicationStarted triggers when all the hosted services have started. And there is my problem, because the collector hosted service should not start until the webhost hosted service has started (the IIS problem), so I decided triggering the collection in the Register callback because I can't move our hosted service to the webhost configure services.

The following code is working ok and there is not further need to check if the host has addresses reported because they are always available when the callback triggers

Do you think this is correct?

```csharp
  private Task ExecuteAsync(CancellationToken cancellationToken)
  {
      _lifetime.ApplicationStarted.Register(async () => await Collect(cancellationToken));
      return Task.CompletedTask;
   }
```

rynowak: That's what I expected you would do, sorry if my example was unclear.

CarlosLanderas: As always, thank you very much @rynowak. Much appreciated!

PR #340, opened 2019-11-18
Fix KeyNotFoundException, in case of the app monitored is down,

Wicusiu:


src/HealthChecks.UI/Core/HostedService/HealthCheckReportCollector.cs
```diff
@@ -197,17 +197,20 @@ private void SaveExecutionHistoryEntries(UIHealthReport healthReport, HealthChec
 
             foreach (var item in execution.Entries)
             {
-                var reportEntry = healthReport.Entries[item.Name];
+                // If the health service is down, no entry in dictionary
+                if(healthReport.Entries.ContainsKey(item.Name)) {
```
sungam3r:
```suggestion
                if (healthReport.Entries.ContainsKey(item.Name))
                { 
```

unaizorrilla:
Merged

Thanks for contribute, a new package 3.0.4 will be available when release build finish

Issue #324, opened 2019-11-07
[RabbitMQ] Health Check fails for SSL/TLS connections


rwkarg:
This commit https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/commit/3c149151d3ccb044995e1df9526e702da7ba744d removes passing in the `_sslOptions`, even if they are specified. This prevents a health check from being successful to an SSL/TLS endpoint on a broker.

Behavior is as expected in version 3.0.0, prior to this commit.

unaizorrilla:
Hi @rwkarg 

You can specify ssloption on ConnectionFactory definition. This works for you?
```

        public async Task be_healthy_if_rabbitmq_is_available_using_iconnectionfactory()
        {
            var connectionString = @"amqp://localhost:5672";

            var factory = new ConnectionFactory()
            {
                Uri = new Uri(connectionString),
                AutomaticRecoveryEnabled = true,
                Ssl= new SslOption(serverName: "localhost", enabled: false)
            };

            var webHostBuilder = new WebHostBuilder()
            .UseStartup<DefaultStartup>()
            .ConfigureServices(services =>
            {
                services
                    .AddHealthChecks()
                    .AddRabbitMQ(sp=>factory,  tags: new string[] { "rabbitmq" });
            })
            .Configure(app =>
            {
                app.UseHealthChecks("/health", new HealthCheckOptions()
                {
                    Predicate = r => r.Tags.Contains("rabbitmq")
                });
            });

            var server = new TestServer(webHostBuilder);

            var response = await server.CreateRequest($"/health")
                .GetAsync();

            response.StatusCode
                .Should().Be(HttpStatusCode.OK);
        }
```

rwkarg:
This constructor:
https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/3c149151d3ccb044995e1df9526e702da7ba744d/src/HealthChecks.RabbitMQ/RabbitMQHealthCheck.cs#L17 
allows passing of SslOptions which get saved internally, but the previously mentioned PR removes any usage of that option.
Behavior in 3.0.0 would use the specified SslOption which is not surprising. The 3.0.1 package will accept the SslOption in the constructor but then ignore it which is surprising.

rwkarg:
[...]

PR #347, opened 2019-11-26
[RabbitMQ] Using SslOptions if passed in

rwkarg:
Addresses #324 
This change will use the IConnectionFactory as configured or the generated ConfigurationFactory with specified SslOption set when creating a connection.

### DIFF
--- a/src/HealthChecks.RabbitMQ/RabbitMQHealthCheck.cs
+++ b/src/HealthChecks.RabbitMQ/RabbitMQHealthCheck.cs
@@ -1,7 +1,6 @@
 ﻿using Microsoft.Extensions.Diagnostics.HealthChecks;
 using RabbitMQ.Client;
 using System;
-using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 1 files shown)


rwkarg:
@unaizorrilla

unaizorrilla:
Hi @rwkarg 
PR is merged !! thanks for contribute on this!

Release build is on appveyor, when finished, a new package with version 3.0.2 will be on NuGet

bkaid:
Upgrading from 3.0.1 to 3.0.2 without SSL turned now passes a null SslOption object and causes a null reference exception in the RabbitMQ client library, when previously a `new SslOption()` ended up getting passed by virtue of not setting ConnectionFactory.Ssl.  Probably doing
```
Ssl = sslOption ?? new SslOption()
```
would have been better to not cause breaking changes and forcing an SslOption to be passed in.

unaizorrilla:
Hi @bkaid 

Thanks for catch this, i pushed a new commit to solve this issue, when release build finished a new package 3.0.3 will be available and solve this issue.
thanks

bkaid:
Confirmed that 3.0.3 fixed it, thanks!

Issue #362, opened 2019-12-13
Invalid package name in root README.md


ahryshchanka:
file:
`Install-Package AspNetCore.HealthChecks.AzureIoTHub`
nuget.org:
`Install-Package AspNetCore.HealthChecks.Azure.IoTHub -Version 3.0.0`

difference: AzureIoTHub and Azure**.**IoTHub, dot in name

PR #363, opened 2019-12-13
fix Azure IotHub package name in readme.md

ahryshchanka:
PR for #362


CarlosLanderas:
Thanks for the fix @ahryshchanka

PR #369, opened 2019-12-17
[Azure key vault] Add certificates with configurable expiry check

CarlosLanderas:
This PR adds the ability to configure azure key vault certificates with an optional "checkExpired" parameter that will throw an expection when the remote certificate has expired.

Sample:

```csharp

services
  .AddHealthChecks()
  .AddAzureKeyVault(setup =>
   {
     setup
        .UseKeyVaultUrl("https://landevault.vault.azure.net")
        .AddSecret("cosmos-key")
        .AddKey("Key1")
        .AddCertificate("landecert")
        .AddCertificate("Cert2", checkExpired: true)
        .UseClientSecrets("client", "secret");
   }
```

### DIFF
--- a/src/HealthChecks.AzureKeyVault/AzureKeyVaultOptions.cs
+++ b/src/HealthChecks.AzureKeyVault/AzureKeyVaultOptions.cs
@@ -22,6 +22,13 @@ namespace HealthChecks.AzureKeyVault
             get { return _keys; }
         }
 
+        internal List<(string, bool)> _certificates = new List<(string key, bool checkExpired)>();
+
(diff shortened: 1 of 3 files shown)


src/HealthChecks.AzureKeyVault/AzureKeyVaultOptions.cs
```diff
@@ -22,6 +22,13 @@ internal IEnumerable<string> Keys
             get { return _keys; }
         }
 
+        internal List<(string, bool)> _certificates = new List<(string key, bool checkExpired)>();
```
sungam3r: private?

PR #382, opened 2019-12-23
Be able to setup KafkaHealthCheck as singleton to re-use producer connections to broker

rguyton:
My team would like to be able to register this health check as a singleton and not initiate a new connection to our broker every time the health is checked.  The following change would allow for using the health-check as-is today and allow others to register the health check as a singleton to re-use connections.

### DIFF
--- a/src/HealthChecks.Kafka/KafkaHealthCheck.cs
+++ b/src/HealthChecks.Kafka/KafkaHealthCheck.cs
@@ -12,7 +12,7 @@ namespace HealthChecks.Kafka
 
         public KafkaHealthCheck(ProducerConfig configuration)
         {
-            if(configuration == null) 
+            if (configuration == null) 
                 throw new ArgumentNullException(nameof(configuration));
             
             _producer = new ProducerBuilder<string, string>(configuration).Build();
--- a/src/HealthChecks.Kafka/KafkaHealthCheck.cs
+++ b/src/HealthChecks.Kafka/KafkaHealthCheck.cs
@@ -8,24 +8,27 @@ namespace HealthChecks.Kafka
 {
     public class KafkaHealthCheck : IHealthCheck
     {
-        private readonly ProducerConfig _configuration;
+        private readonly IProducer<string, string> _producer;
+
         public KafkaHealthCheck(ProducerConfig configuration)
         {
-            _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+            if(configuration == null) 
+                throw new ArgumentNullException(nameof(configuration));
+            
+            _producer = new ProducerBuilder<string, string>(configuration).Build();
         }
+
         public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
         {
             try
             {
-                using (var producer = new ProducerBuilder<string, string>(_configuration).Build())
-                {
-                    var message = new Message<string, string>()
+                var message = new Message<string, string>()
                     {
-                        Key = "healtcheck-key",
+                        Key = "healthcheck-key",
                         Value = $"Check Kafka healthy on {DateTime.UtcNow}"
                     };
 
-                    var result = await producer.ProduceAsync("healthchecks-topic", message);
+                    var result = await _producer.ProduceAsync("healthchecks-topic", message);
 
                     if (result.Status == PersistenceStatus.NotPersisted)
                     {
@@ -33,7 +36,6 @@ namespace HealthChecks.Kafka
                     }
 
                     return HealthCheckResult.Healthy();
-                }
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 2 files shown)


src/HealthChecks.Kafka/KafkaHealthCheck.cs
```diff
@@ -8,32 +8,34 @@ namespace HealthChecks.Kafka
...
-        private readonly ProducerConfig _configuration;
+        private readonly IProducer<string, string> _producer;
+
         public KafkaHealthCheck(ProducerConfig configuration)
         {
-            _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+            if(configuration == null) 
+                throw new ArgumentNullException(nameof(configuration));
+            
+            _producer = new ProducerBuilder<string, string>(configuration).Build();
```
sungam3r: So `KafkaHealthCheck` became `IDisposable`

src/HealthChecks.Kafka/KafkaHealthCheck.cs
```diff
@@ -8,32 +8,34 @@ namespace HealthChecks.Kafka
 {
     public class KafkaHealthCheck : IHealthCheck
     {
-        private readonly ProducerConfig _configuration;
+        private readonly IProducer<string, string> _producer;
+
         public KafkaHealthCheck(ProducerConfig configuration)
         {
-            _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+            if(configuration == null)
```
sungam3r:
```suggestion
            if (configuration == null) 
```

src/HealthChecks.Kafka/KafkaHealthCheck.cs
```diff
@@ -8,32 +8,34 @@ namespace HealthChecks.Kafka
 {
     public class KafkaHealthCheck : IHealthCheck
     {
-        private readonly ProducerConfig _configuration;
+        private readonly IProducer<string, string> _producer;
+
         public KafkaHealthCheck(ProducerConfig configuration)
         {
-            _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
```
unaizorrilla: Hi @rguyton 

We prefer throw expressions instead the if(null) throw, in this case using discards

unaizorrilla: `_ = configuration ?? throw new ArgumentNullException(nameof(configuration));`

src/HealthChecks.Kafka/KafkaHealthCheck.cs
```diff
@@ -8,32 +8,34 @@ namespace HealthChecks.Kafka
 {
     public class KafkaHealthCheck : IHealthCheck
     {
-        private readonly ProducerConfig _configuration;
+        private readonly IProducer<string, string> _producer;
```
unaizorrilla: Why this enable re-use producer connection? Each time healch check is executed a new instance of KafkaHealthCheck is created. If you need singleton you need to store this producer configuration as static ( like azure Service Bus health check do )

My question is, IProducer is intended to be singleton? If true you can do similar things that Azure Service Bus HealthCheck do

rguyton: You can register the Healthcheck itself as a singleton. (see https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-3.1) which will allow the IProducer to be reused.  I'll take a look at the Azure Service Bus HealthCheck.  Thanks for the comments and feedback.

```C#
public class Startup 
{
  ...
  public void ConfigureServices(IServiceCollection collection)
  {
      services.AddSingleton<KafkaHealthCheck>();
      services.AddSingleton(myProducerConfigurationObject);
      ...
      services.AddHealthChecks()
         .AddCheck<KafkaHealthCheck>("kafka");
  }
  ...
}
```

unaizorrilla: Hi @rguyton

You can do that, but, how about register multiple Kafka healtcheck for different Kafka's? In Azure Service Bus HealthCheck you can check different services (topics) and  reusing configuration for each.
Probably is not common have different Kafka healthchecks in the same service. I probably accept this PR but what do you thing about my comments?

rguyton: I think I want to revisit this and check out more in depth what the Azure Service Bus checks are doing.  I have not had a lot of time this past week.

The more I think about it, I'm less than thrilled about removing the dispose. I do have a smaller change I'd like to try to get your way and that would be to make the health check topic configurable.  I think I'll close this pull request for now and try again later.

unaizorrilla: Hi @rguyton 

I think you can modify the IHealthCheckBuilder extension and change sp=>new KafkaHealtchCheck by new KafkaHealthCheck and ensure one singleton instance by kafka registration. This sounds good? can you send a new Pr witth this changes?

PR #385, opened 2019-12-26
[UI Client] Async serialization, http context writer pipeline usage and reduced allocations

CarlosLanderas:
- Async json serialization to prevent thread blocking in UI Client.
- Use asp.net core 3.0 http context pipeline body writer with ` ReadOnlyMemory<byte>` implicit operator
- Reduced string allocations in empty response string and sync json serialization

PR #386, opened 2019-12-26
heap allocation optimization for JsonSerializerOptions

sungam3r:
Relates to #385 

@CarlosLanderas It seems to me that it will be much better and the advantage in comparison with the old sync version will increase more.

### DIFF
--- a/src/HealthChecks.UI.Client/UIResponseWriter.cs
+++ b/src/HealthChecks.UI.Client/UIResponseWriter.cs
@@ -1,20 +1,20 @@
 ﻿using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Diagnostics.HealthChecks;
 using System;
-using System.Threading.Tasks;
+using System.IO;
 using System.Text.Json;
 using System.Text.Json.Serialization;
-using System.IO;
+using System.Threading.Tasks;
 
 namespace HealthChecks.UI.Client
 {
     public static class UIResponseWriter
     {
         private static byte[] emptyResponse = new byte[] { (byte)'{', (byte)'}' };
-        private static JsonSerializerOptions defaultOptions = CreateOptions(null);
+        private static Lazy<JsonSerializerOptions> options = new Lazy<JsonSerializerOptions>(() => CreateOptions());
         const string DEFAULT_CONTENT_TYPE = "application/json";
 
-        private static JsonSerializerOptions CreateOptions(Action<JsonSerializerOptions> jsonConfigurator)
+        private static JsonSerializerOptions CreateOptions()
         {
             var options = new JsonSerializerOptions()
             {
@@ -22,8 +22,6 @@ namespace HealthChecks.UI.Client
                 PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
                 IgnoreNullValues = true,
             };
-
-            jsonConfigurator?.Invoke(options);
 
             options.Converters.Add(new JsonStringEnumConverter());
 
@@ -34,16 +32,10 @@ namespace HealthChecks.UI.Client
             return options;
         }
 
-        private static JsonSerializerOptions TryCreateOptions(Action<JsonSerializerOptions> jsonConfigurator) => jsonConfigurator == null ? defaultOptions : CreateOptions(jsonConfigurator);
-
-        public static async Task WriteHealthCheckUIResponse(HttpContext httpContext, HealthReport report) => await WriteHealthCheckUIResponse(httpContext, report, null);
-
-        public static async Task WriteHealthCheckUIResponse(HttpContext httpContext, HealthReport report, Action<JsonSerializerOptions> jsonConfigurator)
+        public static async Task WriteHealthCheckUIResponse(HttpContext httpContext, HealthReport report)
         {
             if (report != null)
             {
-                var options = TryCreateOptions(jsonConfigurator);
-
                 httpContext.Response.ContentType = DEFAULT_CONTENT_TYPE;
 
                 var uiReport = UIHealthReport
@@ -51,7 +43,7 @@ namespace HealthChecks.UI.Client
 
                 using var responseStream = new MemoryStream();
 
-                await JsonSerializer.SerializeAsync(responseStream, uiReport, options);
+                await JsonSerializer.SerializeAsync(responseStream, uiReport, options.Value);
                 await httpContext.Response.BodyWriter.WriteAsync(responseStream.ToArray());
             }
             else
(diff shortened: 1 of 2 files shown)


CarlosLanderas:
@sungam3r, thanks for trying to improve the Response writer even more. But the signature you need to provide to the ResponseWriter is defined by Microsoft as:

public Func<HttpContext, HealthReport, Task> ResponseWriter { get; set; }

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/samples/HealthChecks.UI.Branding/Startup.cs#L61

The only way I see something like this being done would be wrapping the functin:

```csharp

 var options = new JsonSerializerOptions(); 
//whatever config in JsonSerializer..

.UseEndpoints(config =>
               {
                   config.MapHealthChecks("/health-random", new HealthCheckOptions
                   {
                       Predicate = r => r.Tags.Contains("random"),
                       ResponseWriter = async(context, report) =>
                       {
                          //Wrap the new implementation that caches the options
                           await UIResponseWriter.WriteHealthCheckUIResponse(context, report, options);
                       }
                   });
```

sungam3r:
So I did not change the signature.

CarlosLanderas:
So your idea is that:

sungam3r:
No. Why?

sungam3r:
I just cached the settings object, which is used constantly if no configuration delegate is set.

CarlosLanderas:
Why should that method be public? It is call

CarlosLanderas:
Why should that method be public? It is call

CarlosLanderas:
To be honest, I do not know why that second method is public when the ResponseWriter is automatically invoked by Microsoft Healthchecks, and it also does not make sense to use that Action<JsonSerializerOptions>. I know that was already there...

For me, the best implementation would be this:

```csharp
   public static class UIResponseWriter
    {
        private static byte[] emptyResponse = new byte[] { (byte)'{', (byte)'}' };
        private static Lazy<JsonSerializerOptions> options = new Lazy<JsonSerializerOptions>(() => CreateOptions());

        const string DEFAULT_CONTENT_TYPE = "application/json";

        public static async Task WriteHealthCheckUIResponse(HttpContext httpContext, HealthReport report) => await WriteResponse(httpContext, report);

        private static async Task WriteResponse(HttpContext httpContext, HealthReport report)
        {
            if (report != null)
            {
                httpContext.Response.ContentType = DEFAULT_CONTENT_TYPE;

                var uiReport = UIHealthReport
                    .CreateFrom(report);

                using var responseStream = new MemoryStream();

                await JsonSerializer.SerializeAsync(responseStream, uiReport, options.Value);
                await httpContext.Response.BodyWriter.WriteAsync(responseStream.ToArray());
            }
            else
            {
                await httpContext.Response.BodyWriter.WriteAsync(emptyResponse);
            }
        }

        private static JsonSerializerOptions CreateOptions()
        {
            var options = new JsonSerializerOptions()
            {
                AllowTrailingCommas = true,
                PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
                IgnoreNullValues = true,
            };

            options.Converters.Add(new JsonStringEnumConverter());

            //for compatibility with older UI versions ( <3.0 ) we arrange
            //timespan serialization as s
            options.Converters.Add(new TimeSpanConverter());

            return options;
        }
    }
```

CarlosLanderas:
Indeed, you can get rid of the second private method... Not necessary

sungam3r:
As you wish.

sungam3r:
But wait...

```c#
 .UseHealthChecks("/healthz", new HealthCheckOptions
                {
                    Predicate = _ => true,
                    ResponseWriter = MyDelegate
                })
...
private static Task MyDelegate(HttpContext httpContext, HealthReport report)
        {
            return UIResponseWriter.WriteHealthCheckUIResponse(httpContext, report, opt => opt.Converters.Add(new MyConverter())); // whatever
        }
```

CarlosLanderas:
Updated sample

sungam3r:
I do not really understand. So to change just one serializer setting I need to copy-paste all `UIResponseWriter` default implementation? Do you mean that?

CarlosLanderas:
No. I mean, no JsonSerializerOptions should be configurable from outside (public API). It should follow a fixed non configurable format to work with the react SPA UI.

sungam3r:
Now I understand your point.

Regarding this:
> Allowing users to configure the Json Settings might break the render in the browser
and that:
> The usage should continue being
ResponseWriter = UIResponseWriter.WriteHealthCheckUIResponse

Honestly, you already allow the user to set the entire delegate, so they can always break something for UI anyway. So what's the difference?

CarlosLanderas:
The user is free to use whatever response writer they want of course, but if they want to use our UI, they must use our UI client method as delegate and that is shown in the docs. The react spa knows how to render that response but not custom ones for obvious reasons. 

That's why this method is sealed from external configurations

CarlosLanderas:
For example, the default Microsoft response writer just writes "Healthy" or "Unhealthy" so we rolled out our own UI version that is coupled with out UI front-end

sungam3r:
OK, the general meaning is clear, the rest is not so important. The main thing is that code has become better.

CarlosLanderas:
Thanks for the contribution @sungam3r

cjbush:
I contributed the original method that allowed you to configure the JSON serializer. The reason it was added (and was being used) was to allow you to format your healthcheck names with something other than the default formatting (e.g. if I want my healthcheck name to be "My Database Connection" instead of "my Database Connection"). Removing the method is a breaking change and should cause a major version bump. I just installed the latest package and now all of my configurations are broken, so I'll have to roll back.

sungam3r:
> Removing the method is a breaking change

It was obvious enough, but I did not insist and agreed with @CarlosLanderas to remove this method.

unaizorrilla:
This is a breaking change and probably we need to restore old way and deprecate current package.

@CarlosLanderas can you revert this?

CarlosLanderas:
Sure, we will stay with the initial async version with less allocations I merged and restore the previous signatures.

We will change the behaviour in the future to allow the users transforming de output but not configuring the serializer settings directly as it might break the UI

sungam3r:
Will I need to redo my initial changes?

CarlosLanderas:
@sungam3r propose them in a new PR if you want 👍. I remember the initial version with settings caching and old signatures. That should work.

cjbush:
@CarlosLanderas It's ok. You make a good point about breaking things by configuring the serializer too much. If you want, I can submit another patch that deprecates the current method that allows that and adds another method that takes a settings object with whatever settings you think are appropriate. Something like:

```csharp
public class HealthcheckDisplayOptions
{
    public bool UseCamelCase { get; set; } //or an enum or something
}

public static async Task WriteHealthCheckUIResponse(HttpContext httpContext, HealthReport report, Action<HealthcheckDisplayOptions> config)
{ 
/* Handles configuring the serializer based on the options passed in */ 
}
```

CarlosLanderas:
Sounds good 👍. That DisplayOptions object is good to add new features that users might need and configure them internally without having the change of accidentally clear the necessary converters or mandatory configuration for the UI

sungam3r:
So will there be a rollback or not (another patch)?

Issue #351, opened 2019-12-03
Critical: AddHealthChecksUI calls BuildServiceProvider, causing Singletons to be instantiated multiple times


ffMathy:
AddHealthChecksUI calls BuildServiceProvider, causing Singletons to be instantiated multiple times. This also causes memory issues, and should be fixed.

ffMathy:
Related StackOverflow question: https://stackoverflow.com/questions/48856483/net-core-singleton-creation-is-called-multiple-times

unaizorrilla:
Hi @ffMathy 

Let me check this and I try to respond on this question ASAP!

unaizorrilla:
Hi @ffMathy 

I aggre that call BuildServiceProvider can cause that singletons can be instantiated multiple times, but in our case we use this call to resolve IConfiguration inside AddHealthChecksUI. If you test it, you can verify that the instance resolved for IConfiguration is the same one on Startup and not duplication exist in this case.

I close the issue, reopen is necessary!

Thanks for contribute on this!

ffMathy:
It needs to be reopened.

The problem is not the configuration instance. The problem is that developers assume that their Singleton services run only once.

By using your project, all my Singleton services all of a sudden run twice, leading to memory issues and locking issues.

Microsoft has very clearly defined that this is a bad idea.

If you need the configuration instance, you can access it without calling Build service provider easily. Just turn your AddSingleton call into a callback.

unaizorrilla:
Hi @ffMathy 

Yes, i can access to configuration object without build service provider again, true, but this is a breaking change! 

Meanwhile if you have a problem with some singleton registration you can move AddX after AddHealtchChecksUI.

 ie, agree that we need to fix this, but we need make a breaking change, planned for next version!

ffMathy:
Can you elaborate on how this is a breaking change? I'm fairly confident it can be done with no breaking changes.

CarlosLanderas:
Fixed in version 3.1.1-preview1. No more BuildServiceProvider in ConfigureServices

PR #387, opened 2019-12-30
Non-breaking changes towards fixing issue #351

ffMathy:
This is a little step in the right direction in terms of #351. It is not a breaking change, and it still doesn't fully resolve the issue, but it reduces the `BuildServiceProvider` calls by one at least.

I will make an additional PR later for the rest.


ffMathy:
This can be merged without breaking changes. I'll make an additional commit that will show how we can get the rest of the way.

CarlosLanderas:
@ffMathy thanks for the contribution!. Please, can I ask you to merge master into this branch as some changes happen inside ServiceCollectionExtensions while reviewing previous PR's?

PR #390, opened 2019-12-31
Breaking changes on top of #387 to fully resolve #351

### LINKED ISSUE #351: Critical: AddHealthChecksUI calls BuildServiceProvider, causing Singletons to be instantiated multiple times

ffMathy:
Builds on top of #387 to fully resolve #351. However, this PR introduces potentially breaking changes, while #387 can safely be merged as-is.

### DIFF
--- a/src/HealthChecks.UI/Core/Data/HealthChecksDb.cs
+++ b/src/HealthChecks.UI/Core/Data/HealthChecksDb.cs
@@ -1,5 +1,9 @@
-﻿using HealthChecks.UI.Core.Data.Configuration;
+﻿using System.Linq;
+using HealthChecks.UI.Configuration;
+using HealthChecks.UI.Core.Data.Configuration;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.Extensions.Options;
 
 namespace HealthChecks.UI.Core.Data
 {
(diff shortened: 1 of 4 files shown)


ffMathy:
Like I mentioned, these are breaking changes. Refer to the other PR for the non-breaking changes, and please review that as well, because that could be merged right away. 

So yes, Kubernetes is always registered here (breaking change).

The code to delete and recreate the database is no longer there. It has been replaced with "seeding" (see the changes in here: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/390/files#diff-734256026a936fc4ca53002df74c716eR41-R57).

This does mean that the database does not get recreated, but it will still have the same data. If dropping the database is really a requirement, I can easily re-add it?

ffMathy:
This is the PR that doesn't introduce breaking changes: #387.

unaizorrilla:
Well, I try to checkl #387 first and get this after!

CarlosLanderas:
@ffMathy , I merge the the PR in series but now I am fully refactoring ConfigureServices because we are adding support for several database providers and there will be no BuildServiceProvider in next 
version.

Branch:

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/tree/storage-providers

ffMathy:
Perfect. We should add a test that makes sure all singletons are only instantiated once. Should be fairly easy to make.

CarlosLanderas:
@ffMathy as we talked, this is already fixed in ui 3.1.1-preview1 that also includes new storage providers. Thank you very much for your help as I implemented the hosted service idea :)

ffMathy:
I am aware that it has been fixed, but is there a unit test to ensure it won't happen again?

CarlosLanderas:
There is no call to BuildServiceProvider in all the solution. Just in tests projects. However we can try to add some tests

PR #414, opened 2020-02-12
[docker-image] Extend docs to mention who the default port binding ca…

lanwin:
…n be changed

In my case the docker image binds default to IP6 [::]:80 but I needed the port binding to watch in IP6. So a comment how to change that would have saved me a lot of time.


CarlosLanderas:
Hello @lanwin. This is aspnetcore behaviour and it's documented in Microsoft docs. Not something specific to HealthChecks image. 

cc @unaizorrilla

lanwin:
> Hello @lanwin. This is aspnetcore behaviour and it's documented in Microsoft docs. Not something specific to HealthChecks image.
> 
> cc @unaizorrilla

Yes that is absolutly true. But Its very hard to find in the docs if you dont know where it is. So a mention would help users to get read faster.

unaizorrilla:
Hi @lanwin 

First of all, thanks for contribute on this project. I close the PR because I think this is a behavior documented already on Microsoft docs,

PR #418, opened 2020-02-13
Kubernetes HealthChecks Operator

CarlosLanderas:
Version 3.0.0-beta.1 of the new Kubernetes Healthchecks Operator, targetting netcoreapp3.0

- [ ] Pending documentation

Features:

- [✓] New Custom resource definition  (HealthChecks) (Might change in final version)

- [✓] Automatic UI image deployment / removal and service deployment / removal when creating or removing healthcheck crds.

- [✓] Automatic service discovery, removal and updates using a push model instead of the current UI 
      pull  model. The operator will listen to configured labels in CRD definition and this can be overriden 
      with labels configured in services.
     
- [✓] Automatic UI push endpoint exposure protected by an api key stored and created as a secret by kubernetes operator

Notes:

- The UI Collector has been updated to support updated paths and schemes and reflect the execution histories properly. This was not supported before and now it can be updated by the operator

- The push endpoint can only be enabled by setting the proper configuration and the api key secret is only known by the operator that retrieves the value from a namespaced secret


deploy/operator/crd/healthcheck-crd.yaml
```diff
@@ -0,0 +1,46 @@
...
+  scope: Namespaced
+
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            name:
+              type: string
+            portType:
```
ggmaresca: `serviceType` is a better field name - the type is on the service, not on the port. Defaulting to `ClusterIP`, like a Service does, also seems like a better choice than making it required.

Also,

``` yaml
enum:
- ClusterIP
- LoadBalancer
- NodePort
```

is better for defining a CRD field than a pattern.

deploy/operator/crd/healthcheck-crd.yaml
```diff
@@ -0,0 +1,46 @@
...
+              type: string
+              pattern: ^(LoadBalancer|ClusterIP|NodePort)$
+            portNumber:
+              type: number
```
ggmaresca: Defaulting to `80` seems like an intelligent choice for the port number.

deploy/operator/crd/healthcheck-crd.yaml
```diff
@@ -0,0 +1,46 @@
...
+            uiPath:
+              type: string
+            servicesLabel:
+              type: string
+            servicesHealthPathLabel:
```
ggmaresca: I had made these annotation only (and also added an annotation for the port) in the discovery service.

src/HealthChecks.UI.K8s.Operator/Handlers/DeploymentHandler.cs
```diff
@@ -0,0 +1,127 @@
...
+                _logger.LogError("Error deleting deployment for hc resource: {name} - err: {error}", resource.Spec.Name, ex.Message);
+            }
+        }
+
+        public V1Deployment Build(HealthCheckResource resource)
+        {
+
+            var metadata = new V1ObjectMeta
+            {
+                Labels = new Dictionary<string, string>
```
ggmaresca: It would be better to set `V1ObjectMeta.OwnerReferences` to point to the `HealthCheckResource` instead of relying on labels to find the deployments and services.

src/HealthChecks.UI.K8s.Operator/Handlers/DeploymentHandler.cs
```diff
@@ -0,0 +1,127 @@
...
+                            ["app"] = resource.Spec.Name
+                        }
+                    },
+                    Spec = new V1PodSpec
+                    {
+                        Containers = new List<V1Container>
+                        {
+                            new V1Container
+                            {
+                                ImagePullPolicy = "Never",
```
ggmaresca: Never?

src/HealthChecks.UI.K8s.Operator/Handlers/DeploymentHandler.cs
```diff
@@ -0,0 +1,127 @@
...
+                                Name = Constants.PodName,
+                                Image = Constants.DockerImage,
```
ggmaresca: The image should be configurable on the CRD.

build/dependencies.props
```diff
@@ -72,6 +73,7 @@
     <KubernetesClient>1.5.25</KubernetesClient>
     <DogStatsDCSharpClient>3.3.0</DogStatsDCSharpClient>
     <MicrosoftAzureDevices>1.18.1</MicrosoftAzureDevices>
+    <KubernetesClient>1.6.10</KubernetesClient>
```
ggmaresca: This value is duplicated.

deploy/operator/crd/healthcheck-crd.yaml
CarlosLanderas: Agree

CarlosLanderas:
@ggmaresca , thanks for the review. Let's work to addressd it :)

src/HealthChecks.UI.K8s.Operator/Handlers/DeploymentHandler.cs
CarlosLanderas: This should be "Always", I forgot to change it when doing minikube local testing as the docker image is not published yet.

deploy/operator/crd/healthcheck-crd.yaml
CarlosLanderas: As we merged yesterday your changes, I should adapt this to match the discovery service. I'll turn this into annotations

src/HealthChecks.UI.K8s.Operator/Handlers/DeploymentHandler.cs
CarlosLanderas: Yes, definitely better idea, let me change it

src/HealthChecks.UI.K8s.Operator/Handlers/DeploymentHandler.cs
CarlosLanderas: Why the image should be configurable on the CRD? There is only one image... The docker healthchecks UI image

src/HealthChecks.UI.K8s.Operator/Handlers/DeploymentHandler.cs
CarlosLanderas: Do you mean like this, right?

```csharp
            var metadata = new V1ObjectMeta
            {
                OwnerReferences = new List<V1OwnerReference>
                {
                    new V1OwnerReference { Uid = resource.Metadata.Uid, ApiVersion = resource.ApiVersion, Kind = resource.Kind }
                },
                Labels = new Dictionary<string, string>
                {
                    ["app"] = resource.Spec.Name
                },
                Name = $"{resource.Spec.Name}-deploy",
                NamespaceProperty = resource.Metadata.NamespaceProperty
            };
```

CarlosLanderas: I do not see any parameter to filter by owner reference UID's in _client.ListNamespacedDeploymentAsync, so this means we need to get all services in namespace and filter

deploy/operator/crd/healthcheck-crd.yaml
CarlosLanderas: Yeah, it fits better. I agree

src/HealthChecks.UI.K8s.Operator/Handlers/DeploymentHandler.cs
ggmaresca: There are many reasons why someone might want a custom image. For example, the Docker images right now will run as root, and fail to run on any cluster with a PodSecurityPolicy that forbids running as root.

CarlosLanderas: A custom image created from our base one I guess. Otherwise the operator won't be able to communicate with the UI

deploy/operator/crd/healthcheck-crd.yaml
CarlosLanderas: Only name and serviceLabel is mandatory now. All the other stuff has internal defaults until 1.17 cluster versions are wide spreaded.

src/HealthChecks.UI.K8s.Operator/Handlers/DeploymentHandler.cs
CarlosLanderas: Exposed a new Image and ImagePullPolicy in the CRD

ggmaresca: A Docker image can define the user to run as without needing a security context. Running `adduser` and then using the directive `USER newusername` will work if its PodSecurityPolicy has a `MustRunAs` runAsUser rule while a `USER 1000` directive in the Dockerfile will work with both `MustRunAs` and `MustRunAsNonRoot` PodSecurityPolicy runAsUser rules.

PR #421, opened 2020-02-22
Add  Application DbContext and parallel collection for HealthChecks.UI

maikebing:
-   The same database is now supported with the application
-    Collect endpoint health check reports in parallel


CarlosLanderas:
@maikebing storage providers have already been added in version 3.1.1-preview1.

Issue #399, opened 2020-01-19
HealthCheck UI not able to pull the details when deployed using Azure Dev Spaces


santoshpatro:
I have submitted a bug at : https://stackoverflow.com/questions/59813525/healthcheck-ui-not-able-to-pull-the-details-when-deployed-using-azure-dev-spaces with all the details

Can anyone help me to know how to fix this issue?

unaizorrilla:
Hi @santoshpatro 

The error seems related with sqlite,

This project works fine on local and not when this is deployed using Azure Dev Spaces?

Are your modify default path or connection string for UI local database?

santoshpatro:
I have not modified path or connection string in this case. I just tried the same parameters both in the POC and the original project.

jlcalderingarcia:
We are having the same issue. It seems the sqlite database gets blocked between two or more operations. We were able to reproduce it by running several integration tests on the Api with the configuration in place. Let me know if you would like to have the code sample to create an empty project with the setup.
Thank you very much,

This is the exception we have on azure in case it might help:

> 2020-02-14T08:29:15.396285828Z   _____                               
2020-02-14T08:29:15.396331829Z   /  _  \ __________ _________   ____  
2020-02-14T08:29:15.396338129Z  /  /_\  \___   /  |  \_  __ \_/ __ \ 
2020-02-14T08:29:15.396342829Z /    |    \/    /|  |  /|  | \/\  ___/ 
2020-02-14T08:29:15.396347529Z \____|__  /_____ \____/ |__|    \___  >
2020-02-14T08:29:15.396351929Z         \/      \/                  \/ 
2020-02-14T08:29:15.396356230Z A P P   S E R V I C E   O N   L I N U X
2020-02-14T08:29:15.396360630Z 
2020-02-14T08:29:15.396364630Z Documentation: http://aka.ms/webapp-linux
2020-02-14T08:29:15.396368730Z Dotnet quickstart: https://aka.ms/dotnet-qs
2020-02-14T08:29:15.396372930Z ASP .NETCore Version: 3.0.0
2020-02-14T08:29:15.396377030Z Note: Any data outside '/home' is not persisted
2020-02-14T08:29:15.537009458Z Running oryx -appPath /home/site/wwwroot -output /opt/startup/startup.sh -defaultAppFilePath /defaulthome/hostingstart/hostingstart.dll     -bindPort 8080 -userStartupCommand '' 
[...]

PR #427, opened 2020-02-26
Ensure only one thread deletes the database #399

jlcalderingarcia:
Ensure the database is deleted only at the begining of the first thread initialization in order to avoid Access Denied exception.
Open issue #399

### DIFF
--- a/src/HealthChecks.UI/ServiceCollectionExtensions.cs
+++ b/src/HealthChecks.UI/ServiceCollectionExtensions.cs
@@ -152,7 +152,6 @@ namespace Microsoft.Extensions.DependencyInjection
             }
         }
 
-
         private static volatile bool isDatabaseMigrated;
         private static SemaphoreSlim migrationSemaphore = new SemaphoreSlim(1);
         private static async Task EnsureMigratedAsync(this HealthChecksDb db)
--- a/src/HealthChecks.UI/ServiceCollectionExtensions.cs
+++ b/src/HealthChecks.UI/ServiceCollectionExtensions.cs
@@ -154,10 +154,10 @@ namespace Microsoft.Extensions.DependencyInjection
 
 
         private static volatile bool isDatabaseMigrated;
-        private static SemaphoreSlim migrationSemaphore = new SemaphoreSlim(1, 1);
+        private static Mutex migrationMutex = new Mutex();
         private static async Task EnsureMigratedAsync(this HealthChecksDb db)
         {
-            await migrationSemaphore.WaitAsync();
+            migrationMutex.WaitOne();
             try
             {
                 if (!isDatabaseMigrated)
@@ -169,7 +169,7 @@ namespace Microsoft.Extensions.DependencyInjection
             }
             finally
             {
-                migrationSemaphore.Release();
+                migrationMutex.ReleaseMutex();
             }
         }
     }
--- a/src/HealthChecks.UI/ServiceCollectionExtensions.cs
+++ b/src/HealthChecks.UI/ServiceCollectionExtensions.cs
@@ -154,10 +154,10 @@ namespace Microsoft.Extensions.DependencyInjection
 
 
         private static volatile bool isDatabaseMigrated;
-        private static Mutex migrationMutex = new Mutex();
+        private static SemaphoreSlim migrationSemaphore = new SemaphoreSlim(1, 1);
         private static async Task EnsureMigratedAsync(this HealthChecksDb db)
         {
-            migrationMutex.WaitOne();
+            await migrationSemaphore.WaitAsync();
             try
             {
                 if (!isDatabaseMigrated)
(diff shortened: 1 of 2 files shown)


src/HealthChecks.UI/ServiceCollectionExtensions.cs
```diff
@@ -114,6 +116,8 @@ public static IServiceCollection AddWebhooksEndpointHttpClient(this IServiceColl
             .Services;
         }
 
+        private static bool databaseDeleted = false;
```
sungam3r:
```suggestion
        private static bool databaseDeleted;

```

jlcalderingarcia: Thank you very much! I didn't realize of it. Also renamed the variable

src/HealthChecks.UI/ServiceCollectionExtensions.cs
```diff
@@ -129,8 +133,14 @@ static async Task CreateDatabase(IServiceProvider serviceProvider)
                 var settings = scope.ServiceProvider
                     .GetService<IOptions<Settings>>();
 
-                await db.Database.EnsureDeletedAsync();
+                deletionMutex.WaitOne();
```
sungam3r: Also you can change sync wait to `SemaphoreSlim.WaitAsync`.

src/HealthChecks.UI/ServiceCollectionExtensions.cs
```diff
@@ -114,6 +115,8 @@ public static IServiceCollection AddWebhooksEndpointHttpClient(this IServiceColl
             .Services;
         }
 
+        private static volatile bool isDatabaseDeleted = false;
```
sungam3r:
```suggestion
        private static volatile bool isDatabaseDeleted;
```

src/HealthChecks.UI/ServiceCollectionExtensions.cs
```diff
@@ -151,5 +151,19 @@ await db.Configurations
...
+        private static SemaphoreSlim deletionSemaphore = new SemaphoreSlim(1);
+        private static async Task EnsureMigratedAsync(this HealthChecksDb db)
+        {
+            await deletionSemaphore.WaitAsync();
+            if (!isDatabaseDeleted)
+            {
+                await db.Database.EnsureDeletedAsync();
+                isDatabaseDeleted = true;
+            }
+            await db.Database.MigrateAsync();
```
sungam3r: Do you want all subsequent threads to execute this line?

jlcalderingarcia: Not really. I read again the code and it is clear what I'm doing now. I think with the latest code I just pushed it will be clear. Thanks!

src/HealthChecks.UI/ServiceCollectionExtensions.cs
```diff
@@ -151,5 +151,24 @@ await db.Configurations
...
+            try
+            {
+                if (!isDatabaseMigrated)
+                {
+                    await db.Database.EnsureDeletedAsync();
+                    await db.Database.MigrateAsync();
+                    isDatabaseMigrated = true;
+                }
+            }
+            finally {
```
sungam3r:
```suggestion
            finally
            {
```

test/FunctionalTests/HealthChecks.UI/ConcurrencyTests.cs
```diff
@@ -0,0 +1,109 @@
...
+    {
+
+        private async Task BaseConcurrencyTest()
+        {
+            var webHostBuilder = new WebHostBuilder()
+              .UseStartup<DefaultStartup>()
+              .ConfigureServices(services =>
+              {
+                  services.AddHealthChecks();
+                  Action<Settings> setupSettings = settings => {
```
sungam3r:
```suggestion
                  Action<Settings> setupSettings = settings =>
                  {
```

jlcalderingarcia:
I changed back to Mutex since using the SemaphoreSlim caused deadlocks and I can't figure it out why

sungam3r:
@jlcalderingarcia 

>  private static SemaphoreSlim migrationSemaphore = new SemaphoreSlim(0, 1);

You initialize your semaphore with an initial count of zero. This means that all who call Wait/WaitAsync will block.

sungam3r:
So, right, it is 100% deadlock.

jlcalderingarcia:
In this case I changes the SemaphoreSlim configuration to start with count 1 and max count 1, and still it is having a deadlock. The only way I found to make it work is using the Mutex. I don't knoe exactly why it is happenning, I've tryied the basic configurations for the Semaphore, using just the initial count 1, having initial and max count 1, having initial count 0 and max count 1 and in all of them it ended on a deadlock. I waited for 15min of build, when it usually takes around 7 and it is still locked. So, I'll rollback to Mutex solution.
I guess if it is not good enought we should close the PR.
Thanks!

PR #433, opened 2020-03-04
S3 to use default credentials when no keys

ivanau:
A change to allow S3HealthCheck to use default credentials when no keys provided.


unaizorrilla:
Hi @ivanau 

Thanks for contributed this! This is merged, we are on progress to downgrade .net standard 2.1 to .net standard 2.0 and a new version package will be published...

Thanks!!

udlose:
@unaizorrilla that is great to hear!  Any approximate idea on when the new package with this functionality would be released?

unaizorrilla:
Now :-) 

https://www.nuget.org/packages/AspNetCore.HealthChecks.Aws.S3/

Issue #415, opened 2020-02-12
[Question] Webhooks [[LIVENESS]] and [[FAILURE]]


CarljohanDanling:
Hi. If I have a collection of healthchecks under one specific HealthCheckEndpoint. Is it possible in the [[LIVENESS]] and [[FAILURE]] to target the specific one that Webhooks notifies Slack about?

For example. If I have 3 apps under the endpoint name "Agents" (app1, app2 & app3). If app1 fails, the [[LIVENESS]] says "Agents are failing...". I would like to point out that just app1 is failing, not the collection of Agents.

Any ideas how to do this?

CarlosLanderas:
Hello @CarljohanDanling. What do you think about being able to customize if the payload should be notified, the message and the descriptions using funcs where you receive the UIReport, something like this:

```csharp
setup
.AddWebhookNotification("webhook1",
uri: "status/200?code=ax3rt56s", payload: "{ message: \"Webhook report for [[LIVENESS]]: [[FAILURE]]\"}",
restorePayload: "{ message: \"[[LIVENESS]] is back to life\"}",
shouldNotifyFunc: report => DateTime.UtcNow.Hour > 8 && DateTime.UtcNow.Hour < 20
customMessageFunc: (report) =>
{
    var failing = report.Entries.Where(e => e.Value.Status == UIHealthStatus.Unhealthy);
    return $"{failing.Count()} healthchecks are failing";
}, 
customDescriptionFunc: report =>
{
    var failing = report.Entries.Where(e => e.Value.Status == UIHealthStatus.Unhealthy);
    return $"HealthChecks with names {string.Join("/", failing.Select(f => f.Key))} are failing";
});

```

CarljohanDanling:
That looks good!

CarlosLanderas:
I've already coded that implementation last night. So if everyone finds it interesting we will add this new customization feature

CarlosLanderas:
This feature will be available in UI version 3.0.11

PR #451, opened 2020-03-15
Custom webhooks notification, failure and description user functions #415

CarlosLanderas:
Issue #415 

Several users reported the need of configuring the message that is sent to Microsoft Teams, Slack or any other webhook target.

With this PR,  3 new functions are added to Webhook endpoint configuration.

**ShouldNotify**, receives the UI Report and returns a boolean to let the notifier know if the webhook should be invoked (Useful to configure working hours for example)

**CustomMessageFunc**: Configure the output of [[FAILURE]] bookmark
**CustomDescriptionFunc**: Configure the output of [[DESCRIPTION]] bookmark

Sample:

```csharp
setup
.AddWebhookNotification("webhook1",
    uri: "status/200?code=ax3rt56s", payload: "{ message: \"Webhook report for [[LIVENESS]]: [[FAILURE]] - Description: [[DESCRIPTIONS]]\"}",
    restorePayload: "{ message: \"[[LIVENESS]] is back to life\"}",
    shouldNotifyFunc: report => DateTime.UtcNow.Hour >= 8 && DateTime.UtcNow.Hour <= 23
    customMessageFunc: report =>
    {
        var failing = report.Entries.Where(e => e.Value.Status == UIHealthStatus.Unhealthy);
        return $"{failing.Count()} healthchecks are failing";
    }, 
    customDescriptionFunc: report =>
    {
        var failing = report.Entries.Where(e => e.Value.Status == UIHealthStatus.Unhealthy);
        return $"HealthChecks with names {string.Join("/", failing.Select(f => f.Key))} are failing";
    });
```

- [x] Code compiles correctly
- [x] Extended documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,13 +289,15 @@ Sample:
 
 You can also use relative urls when using IConfiguration providers like appsettings.json
 
-### Failure Notifications
+### Webhooks Failure Notifications
 
 If the **WebHooks** section is configured, HealthCheck-UI automatically posts a new notification into the webhook collection. HealthCheckUI uses a simple replace method for values in the webhook's **Payload** and **RestorePayload** properties. At this moment we support two bookmarks:
 
 [[LIVENESS]] The name of the liveness that returns _Down_.
 
 [[FAILURE]] A detail message with the failure.
+
+[[DESCRIPTIONS]] Failure descriptions
 
 The [web hooks section](./doc/webhooks.md) contains more information and webhooks samples for Microsoft Teams, Azure Functions, Slack and more.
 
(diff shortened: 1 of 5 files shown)


natiox:
We're having problems with our webhook notifications spamming.  Our application has a requirement to only notify upon change in status.  Would it be possible to raise an event upon service restoration or for the restorePayload to throttle itself using the shouldNotifyFunc?  If the latter, it would be nice if UIHealthReport.Entries contained all entries and not just failures.  Alternatively, having a specific "shouldNotifyUponRestoreFunc" would enable handling of the requirement.  Do you have any other approaches we might be missing?

Issue #317, opened 2019-10-31
Allow configuring Elastic Timeout


cypressious:
The Elastic client allows configuring request timeouts via `ConnectionSettings.RequestTimeout`, however this is not exposed in `AspNetCore.HealthChecks.Elasticsearch`.

Consider adding a way to configure the timeout.

unaizorrilla:
PR welcome ! :-)

inkel:
I'm happy to send a PR for this, however, I've got a couple of questions:

Both current extensions `IHealthChecksBuilder.AddElasticsearch` accept an optional, last parameter `TimeSpan? timeout = default`. Now, according to the documentation, this parameter is _An optional System.TimeSpan representing the timeout of the check._

My question would be: should we use that value too and pass it to `ElasticsearchOptions` or should we add a new one `TimeSpan? requestTimeout = default`?

PR #452, opened 2020-03-16
Allow configuring Elastic Search Request Timeout #317

CarlosLanderas:

Allow configuring ElasticSearch client timeout. 

We can use setup.RequestTimeout to configure the request timeout timespan:
```csharp
            services.AddHealthChecks().AddElasticsearch(setup =>
            {
                setup.RequestTimeout = new TimeSpan(0, 0, 6);
            });
```

If no setup.RequestTimeout is configured but the user configured the global healthcheck timeout, we pass this value to the RequestTimeout so the connection times out when the healthcheck does.

```csharp
            services.AddHealthChecks().AddElasticsearch(setup =>
            {
                
            }, timeout: new TimeSpan(0,0,7));
```

If no option is configured, the HealthCheck will stay with it's current behaviour. No client request timeout would be configured.

**Which issue(s) this PR fixes**: #317

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.Elasticsearch/ElasticsearchHealthCheck.cs
+++ b/src/HealthChecks.Elasticsearch/ElasticsearchHealthCheck.cs
@@ -26,6 +26,11 @@ namespace HealthChecks.Elasticsearch
                 {
                     var settings = new ConnectionSettings(new Uri(_options.Uri));
 
+                    if (_options.RequestTimeout.HasValue)
+                    {
+                        settings.RequestTimeout(_options.RequestTimeout.Value);
+                    }
+
                     if (_options.AuthenticateWithBasicCredentials)
                     {
                         settings = settings.BasicAuthentication(_options.UserName, _options.Password);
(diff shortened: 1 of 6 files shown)


src/HealthChecks.Elasticsearch/ElasticsearchHealthCheck.cs
```diff
@@ -26,6 +26,11 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
                 {
                     var settings = new ConnectionSettings(new Uri(_options.Uri));
 
+                    if (_options.RequestTimeout.HasValue)
+                    {
+                        settings.RequestTimeout(_options.RequestTimeout.Value);
```
sungam3r: Why not `settings = settings.RequestTimeout(_options.RequestTimeout.Value);` as all cases below ?

test/UnitTests/ElasticSearchUnitTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            var services = new ServiceCollection();
+            var settings = new ElasticsearchOptions();
+            services.AddHealthChecks().AddElasticsearch(setup =>
+            {
+                settings = setup;
+            });
+
+            settings.RequestTimeout.Should().NotHaveValue();
+
```
sungam3r:
```suggestion
```

Issue #445, opened 2020-03-13
[RabbitMQ] TCP Socket churn when executing health check


rwkarg:
commit https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/commit/2a66bc9588701f7a50d33eb2652727a562ce5f33 reverted the reuse of a single connection when checking health to address the shutdown issue reported in #412 

While the shutdown issue is fixed (which was caused by the IConnectionFactory being passed in requesting Foreground threads for the RabbitMQ socket management), it does now result in TCP socket churn since a new connection is created for every health check evaluation. Eventually this results in ephemeral port exhaustion (given frequent enough health checks) and the process becomes unable to initiate TCP connections.

rwkarg:
Looking at potential resolutions to reintroduce Connection reuse without also reintroducing the undesired shutdown behavior reported in #412 

There doesn't seem to be a direct way to clean up a health check on application shutdown.

Ideally the RabbitMQHealthCheck (or any implementation of IHealthCheck) could be made `IDisposable` and then the framework provided DefaultHealthCheckService would implement `IDisposable` as well and handle necessary disposal of registered health checks. Alternatively the `HealthCheckServiceOptions` could implement `IDisposable` and handle the same clean up.
Unfortunately, that would be an update to .NET and not arrive until .NET 5.0.

For now, it may be worth looking at either plumbing in a generic IDisposable handler trigger by a passed in IApplicationLifetime instance (may be too broad of a solution for this situation? unintended consequences?) or more specific handling of RabbitMQHealthCheck clean up. This would allow the connection reuse to be put back without reintroducing the undesired shutdown behavior reported in #412

rwkarg:
I'm going to take a look at what implementing specific shutdown handling of `RabbitMQHealthCheck` instances in the `HealthCheckServiceOptions.Registrations` might look like.

rwkarg:
One difficulty is that while I can make changes to the extension method (`.AddRabbitMQ`) to wire up the cleanup, there isn't a straight forward way to include such clean up in the example code from #412 where they add the `RabbitMQHealthCheck` outside of the extension method, similar to this:

```
services.AddHealthChecks().Add(
                new HealthCheckRegistration(
                    "rabbitmq",
                    provider => new RabbitMQHealthCheck(provider.GetService<IConnectionFactory>()),
                    HealthStatus.Healthy,
                    new string[0]));

[...]

Issue #412, opened 2020-02-11
[BUG] RabbitMQHealthCheck does not close connection, host hangs on exit.


ig-sinicyn:
**Repro steps:**
Run a host with RabbitMQHealthCheck configured via ConnectionFactory(). Trigger a health check. Press Ctrl-C on a host console.
**Expected behavior:**
Host terminates successfully.
**Actual behavior:**
Host process does not exit even after host shutdown completes.
```cs
	services.AddSingleton<IConnectionFactory>(provider =>
	{
		var connFactory = new ConnectionFactory
		{ ... };
		return connFactory;
	});
	services
		.AddHealthChecks()
		.Add(
			new HealthCheckRegistration(
				"rabbitmq",
				provider => new RabbitMQHealthCheck(provider.GetService<IConnectionFactory>()),
				HealthStatus.Healthy,
				new string[0]));
```
endpoint for checks:
```cs
	app.UseEndpoints(endpoints =>
	{
		endpoints.MapControllers();
		endpoints.MapHealthChecks("/status", new HealthCheckOptions()
		{
			ResponseWriter = (c, r) => c.Response.WriteAsync(JsonSerializer.Serialize(r))
		});
	});
```
launchsettings.json:
```json
"RabbitMqCheckDemo": {
      "commandName": "Project",
      "launchBrowser": true,
      "launchUrl": "status",
      "applicationUrl": "https://localhost:5001;http://localhost:5000",
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Development"
      }
    }
```
Sample: [RabbitMqCheckDemo.zip](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/files/4186035/RabbitMqCheckDemo.zip)

unaizorrilla:
Hi @ig-sinicyn 

I'm fixing the issue, meanwhile you can workaround your code with:

```csharp
              services.AddSingleton<IConnection>(sp =>
             {
                 return new ConnectionFactory()
                 {
                     UserName = "guest",
                     Password = "guest",
                     Port = 5672,
                     HostName = "localhost"
                 }.CreateConnection();
             });

            services.AddHealthChecks()
                .AddRabbitMQ(name: "rabbitmq1");
```

unaizorrilla:
Commiting and a new release build is on AppVeyor, when finish a new package 3.0.5 will be available o NuGet.

[...]

PR #456, opened 2020-03-16
Preventing shutdown hangs while still reusing connections when possible

rwkarg:
Enables connection reuse when possible while avoiding reintroducing the shutdown hang reported in #412

[RabbitMQ] TCP Socket churn when executing health check
 
Please reference the issue this PR will close: #445

Adding an InternalRabbitMQHealthCheck that is only used in cases where we know that the health check instance will be reused instead of creating a new RabbitMQHealthCheck for each health evaluation occurance.

Updating all extension methods to ensure that a single InternalRabbitMQHealthCheck instance is used for each registration instead of having a new instance created for each health evaluation.

No. Attempts were made to avoid any user-facing changes in this PR.
Ideally there would be ways to prevent or mitigate doing the "wrong thing" but that would result in breaking changes from already released APIs. We won't every cause process termination to hang because of references to an IConnection using Foreground IO threads. BUT we do still have some circumstances where users can either request Foreground IO threads to be used which stops us from reusing connections (This will result in connection churn against the RabbitMQ broker and eventually ephemeral port exhaustion given sufficiently frequent health checks) as well as the ability for users to manually (outside of the extension methods) register the public RabbitMQHealthCheck in such a way that a new RabbitMQHealthCheck instance will be created for each health check evaluation which will negate any attempt to reuse a connection inside the RabbitMQHealthCheck (since it's only used once and then thrown away).

Specifically, removing (or making internal) the public RabbitMQHealthCheck would prevent usages such as the `new HealthCheckRegistration` pattern shown in #412.

Additionally, logging a warning (or even strictly preventing) when an IConnectionFactory that is configured to use Foreground IO threads would prevent that use case from being disqualified from connection reuse.

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/src/HealthChecks.RabbitMQ/RabbitMQHealthCheck.cs
+++ b/src/HealthChecks.RabbitMQ/RabbitMQHealthCheck.cs
@@ -83,7 +83,6 @@ namespace HealthChecks.RabbitMQ
                 // benefit from reusing the internally stored connection).
                 else
                 {
-                    
                     if (_rmqConnection == null || !_rmqConnection.IsOpen)
                     {
                         _rmqConnection?.Abort(0);
--- a/src/HealthChecks.RabbitMQ/RabbitMQHealthCheck.cs
+++ b/src/HealthChecks.RabbitMQ/RabbitMQHealthCheck.cs
@@ -11,12 +11,18 @@ namespace HealthChecks.RabbitMQ
     {
         private readonly IConnectionFactory _connectionFactory;
         private IConnection _rmqConnection;
+        protected bool AttemptConnectionReuse = false;
 
         public RabbitMQHealthCheck(string rabbitMqConnectionString, SslOption sslOption = null)
         {
             var connectionFactory = new ConnectionFactory
             {
                 Uri = new Uri(rabbitMqConnectionString ?? throw new ArgumentNullException(nameof(rabbitMqConnectionString))),
+                // Using background threads to ensure that process shutdown isn't waiting for these IO threads to complete.
+                // This addresses issue #412 which reports that the process hangs on shutdown if the connection used
+                // for health checks is using foreground threads and isn't explicitly disposed.
+                UseBackgroundThreadsForIO = true, 
+
                 AutomaticRecoveryEnabled = true // Explicitly setting to ensure this is true (in case the default changes)
             };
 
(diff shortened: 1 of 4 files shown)


rwkarg:
I would imagine if this goes forward that we'll want to also have documentation updated to have guidance on the "right" way to use these to ensure that connections get reused.

rwkarg:
@unaizorrilla Does this seem like a reasonable way to move forward?

src/HealthChecks.RabbitMQ/RabbitMQHealthCheck.cs
```diff
@@ -11,12 +11,18 @@ public class RabbitMQHealthCheck
     {
         private readonly IConnectionFactory _connectionFactory;
         private IConnection _rmqConnection;
+        protected bool AttemptConnectionReuse = false;
```
sungam3r:
```suggestion
        protected bool _attemptConnectionReuse;
```

src/HealthChecks.RabbitMQ/RabbitMQHealthCheck.cs
```diff
@@ -48,12 +54,44 @@ public Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, Canc
...
+                }
+                // We're safe holding on to the connection and not disposing it here because
+                // the factory is configured to use background threads and won't cause the
+                // process shutdown to hang (as would happen with foreground thread that are still running)
+                // and we've been told (through AttemptConnectionReuse) to assume that this is a long
+                // lived RabbitMQHealthCheck instance (so we'll reuse this instance and be able to
+                // benefit from reusing the internally stored connection).
+                else
+                {
+
```
sungam3r:
```suggestion
```

src/HealthChecks.RabbitMQ/RabbitMQHealthCheck.cs
```diff
@@ -48,12 +54,44 @@ public Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, Canc
...
+                //
+                // It is not safe if we don't know that this RabbitMQHealthCheck instance 
+                // may be transient, where a new one created for each new health check evaluation,
+                // since we would then never reuse this instance to reuse the internally stored connection.
+                //
+                // It is also not safe if the IConnectionFactory.UseBackgroundThreadsForIO is
+                // set to false as the foreground threads of an internally stored connection would
+                // prevent the process from terminating since there isn't a hook to dispose
+                // the connection on shutdown.
+                if (!AttemptConnectionReuse)
```
sungam3r: Maybe switch order of if/else branches?

src/HealthChecks.RabbitMQ/RabbitMQHealthCheck.cs
rwkarg: There is one other `protected` field in this repository and it uses PascalCasing:
https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/bc0f481404412015aa11aeb30709eebba5dcd9dc/src/HealthChecks.Publisher.Prometheus/LivenessPrometheusMetrics.cs#L14

Since there is no `.editorconfig` to specify the desired convention, I've just matched the existing precedent.

sungam3r: OK, then `protected bool AttemptConnectionReuse;`

rwkarg:
@unaizorrilla Can you take a look when you get a chance?

unaizorrilla:
sure!

unaizorrilla:
I'm so sorry @rwkarg  but at the end I'm not sure about the problem to solve here and probably this is my fault!

At this moment you can inject IConnection or IConnectionFactory from the ServiceProvider and own this services, I think is not the best idea to do a lot of things, like in this pr, to prevent developer miss configurations!

rwkarg:
Moving discussion to #445

rwkarg:
This issue has been addressed by setting the IConnectionFactory based registrations as Obsolete (to be removed at some future version): https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/commit/6649adc87823979cfca8f050b991b23406a55827

Issue #438, opened 2020-03-05
[UI] No endpoints when using a non-conformant healthcheck endpoint


amccool:
I have several websites with handrolled healthchecks.  their output looks like this:
`{"utcTime":"2020-03-05T23:02:17.5793442Z","machine":"DODAD","version":{"major":3,"minor":1,"build":0,"revision":32,"majorRevision":0,"minorRevision":32}}`

When I add these endpoint to the UI, I get `NullReferenceException` in my debug output, and the UI doesnt render any endpoints at all.

`settings.AddHealthCheckEndpoint("xxx", new Uri("https://olderwebsite/api/healthcheck").ToString());
`

HealthChecks.UI.Core.HostedService.HealthCheckCollectorHostedService: Error: HealthCheck collector HostedService threw an error: Object reference not set to an instance of an object.

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at HealthChecks.UI.Core.Notifications.WebHookFailureNotifier.GetFailedMessageFromContent(UIHealthReport healthReport) in C:\Users\alexm\source\repos\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\Core\Notifications\WebHookFailureNotifier.cs:line 124
   at HealthChecks.UI.Core.Notifications.WebHookFailureNotifier.NotifyDown(String name, UIHealthReport report) in C:\Users\alexm\source\repos\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\Core\Notifications\WebHookFailureNotifier.cs:line 41
   at HealthChecks.UI.Core.HostedService.HealthCheckReportCollector.Collect(CancellationToken cancellationToken) in C:\Users\alexm\source\repos\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\Core\HostedService\HealthCheckReportCollector.cs:line 65
   at HealthChecks.UI.Core.HostedService.HealthCheckCollectorHostedService.Collect(CancellationToken cancellationToken) in C:\Users\alexm\source\repos\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\Core\HostedService\HealthCheckReportCollectorHostedService.cs:line 87
```

UIHealthReport.Entries is null in
```
        private string GetFailedMessageFromContent(UIHealthReport healthReport)
        {
            var failedChecks = healthReport.Entries.Values
[...]

PR #460, opened 2020-03-18
let a 200 be a healthy endpoint

amccool:
when a non-conformant json payload is returned, it doesn't deserialize in a `UIHealthReport`.   Instead a default UIHealthReport is return (with the default of Status = Unhealthy)

Please reference the issue this PR will close: #438

I dont have docker running locally - so my test runs looks like this
```
Test Run Failed.
Total tests: 118
     Passed: 75
     Failed: 43
 Total time: 13.6614 Minutes
```

no

- [x] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/src/HealthChecks.UI/Core/HostedService/HealthCheckReportCollector.cs
+++ b/src/HealthChecks.UI/Core/HostedService/HealthCheckReportCollector.cs
@@ -88,7 +88,22 @@ namespace HealthChecks.UI.Core.HostedService
 
                 var response = await _httpClient.GetAsync(absoluteUri);
 
-                return await response.As<UIHealthReport>();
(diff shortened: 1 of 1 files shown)


CarlosLanderas:
@amccool, as I explained in the issue https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/438#issuecomment-600750667 there are plenty of options to split your healthchecks endpoints following different strategies using tags and paths.

I do not see why the UI should be creating this empty fake reports from non compliant endpoints. Just filter these HealthChecks out of the UI or implement and follow it's response writer.

Don't you think?

amccool:
I sure would like to not have to modify the code for my existing endpoints, but I understand the want for consistency.

PR #463, opened 2020-03-26
Update ConsulHealthCheckBuilderExtensions.cs to allow runtime changes

mphelt:
Small change that will allow consul settings to be changed during runtime. Pretty similar to what happens in NpgSqlHealthCheckBuilderExtensions.AddNpgSql().

- [x] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


unaizorrilla:
Hi @papugamichal 

Can you tell me more about this scenario @papugamichal ?

unaizorrilla:
Can you tell me more about this scenario, a sample of use etc

mphelt:
@unaizorrilla 
I would like to be able to change (in runtime) the path to Consul defined in appsettings.json.
Mostly for test purposes, to simulate Consul downtime.

Current implementation already uses HealthCheckRegistration constructor with health check factory (Func<IServiceProvider, ConsulHealthCheck>) as an argument. The only difference introduced in here is to execute Consul setup (Action<ConsulOptions>) in this factory. This way it will not be executed only once during the registration, but each time health check is to be used.

unaizorrilla:
Hi @mphelt 

Merged on master, a new release build is on appveyor, when finished a new package with version 3.0.1 will be available on NuGet!

PR #466, opened 2020-03-30
Add support for non-URI EventStore connection strings.

sjp:

When attempting to health check a working EventStore connection, I was always getting an `Unhealthy` response.

After debugging, I discovered that there is an assumption in the code that the connection string must be a URI. This is not always the case, and the connection string I was using was similar to the following one taken from the EventStore documentation:

```csharp
var connectionString = "ConnectTo=tcp://admin:changeit@localhost:1113; HeartBeatTimeout=500";
```

I have added support for connection strings in the format shown above.

- [X] Code compiles correctly
- [X] Created/updated tests
- [X] Unit tests passing


unaizorrilla:
Hi @sjp 

LGTM , fix tests and merge. When a release build finished a new package will be on nuget with 3.0.1

PR #471, opened 2020-03-31
[RabbitMQ] Adding Functional Test to ensure connection reuse

rwkarg:
As of this checkin, these two ConnectionFactory tests are failing because connections are not reused

Provides tests to ensure that RabbitMQ connections are reused when registering a health check with an IConnectionFactory (both directly and via DI).

Provides failing cases for ongoing connection reuse work.

No. Just adding functional tests.

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/test/FunctionalTests/HealthChecks.RabbitMQ/RabbitMQHealthCheckTests.cs
+++ b/test/FunctionalTests/HealthChecks.RabbitMQ/RabbitMQHealthCheckTests.cs
@@ -8,6 +8,7 @@ using Microsoft.Extensions.DependencyInjection;
 using RabbitMQ.Client;
 using System;
 using System.Net;
+using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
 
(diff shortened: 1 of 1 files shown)


test/FunctionalTests/HealthChecks.RabbitMQ/RabbitMQHealthCheckTests.cs
```diff
@@ -187,5 +188,120 @@ public async Task be_healthy_if_rabbitmq_is_available_and_specify_default_sslopt
...
+            afterTwoConnectionCount.Should().Be(1, because: "only one connection should be used for health checks");
+        }
+    }
+
+    internal class WrappedConnectionFactory : ConnectionFactory
+    {
+        private int _createdConnections;
+        public int CreatedConnections => _createdConnections;
+
+
```
sungam3r, line 299:
```suggestion
```

unaizorrilla:
Hi @rwkarg 

Closed by #445, we mark this overloads as obsoletes, we have enought ways to register a IConnection, directly or from connection string that reuse connections and does not hang the shutdown app.

PR #468, opened 2020-03-31
Upgrade HealthChecks to netstandard 2.1 and UI and samples to netcoreapp 3.1

CarlosLanderas:
**What this PR does / why we need it**:

Target to latest netstandard and netcoreapp versions.
The new healthchecks nuget versions will be 3.1.X

**Which issue(s) this PR fixes**:

Please reference the issue this PR will close: #467

Please make sure you've completed the relevant tasks for this PR, out of the following list:

- [X] Code compiles correctly
- [X] Unit tests passing
- [X] End-to-end tests passing

PR #485, opened 2020-04-17
Target packages back netstandard2.0 for greater compatibility

CarlosLanderas:

After listening to one of our users we have considered downgrade packages from netstandard2.1 to netstandard 2.0 to increase compatibility surface as we are not using any of the added features included and we do not want to force users to upgrade their solutions.

Although all our healthchecks are already published in version 2.X.X in nuget, people might need the new storage providers that were published under version 2.1

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/468#issuecomment-615240412

Thanks to @vip32 for opening our eyes from a user perspective.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup Label="SDK Versions">
-    <NetStandardTargetVersion>netstandard2.1</NetStandardTargetVersion>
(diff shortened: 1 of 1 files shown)


CarlosLanderas:
@unaizorrilla give me the OK and I republish the packages

PR #528, opened 2020-05-18
Upgrade prometheus-net to v3.x

### LINKED ISSUE #419: Prometheus-Net 3.x is not supported
The AspNetcore.HealthChecks.Publisher.Prometheus package does not support the latest Prometheus-Net versions of 3.x. 

It causes an exception to be thrown in LivenessPrometheusMetrics.cs at 
Registry = DefaultCollectorRegistry.Instance; 
and makes the package unusable.

Daniel15:
Upgrades the Prometheus integration to use v3 of the `prometheus-net` library.

This is the smallest amount of code needed to change in order to upgrade prometheus-net and keep everything in its current state. 

To test it, I enabled the Prometheus metric endpoint in the `HealthChecks.Sample` project.

There's a bit of weirdness in this project's integration into Prometheus - Instead of having its own separate endpoint that updates the metrics and returns them in Prometheus format, it should instead rely on the standard `/metrics` Prometheus endpoint exposed by `endpoints.MapMetrics()` from `prometheus-net.AspNetCore`, and integrate into Prometheus by using `Registry.AddBeforeCollectCallback` (so that Prometheus calls this library to update the metrics, rather than going the other way around). However, that can be done in a separate PR.

Please reference the issue this PR will close: 
Fixes #419

N/A

This is a breaking change as v3 of prometheus-net is not API-compatible with v2. People just using `AspNetCore.Diagnostics.HealthChecks` without any customisations to Prometheus (eg. adding their own metrics) will be fine, but people that have deeper integrations into Prometheus will need to upgrade their own code to use the v3 API.

- [x] Code compiles correctly
- [ ] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -70,7 +70,7 @@
     <NewtonsoftJson>12.0.2</NewtonsoftJson>
     <EventStoreClient>5.0.2</EventStoreClient>
     <MicrosoftApplicationInsights>2.10.0</MicrosoftApplicationInsights>
-    <PrometheusNet>2.1.3</PrometheusNet>
+    <PrometheusNet>3.5.0</PrometheusNet>
     <AWSSKDS3>3.3.29</AWSSKDS3>
     <MicrosoftAzureKeyVault>3.0.4</MicrosoftAzureKeyVault>
     <DogStatsDCSharpClient>3.3.0</DogStatsDCSharpClient>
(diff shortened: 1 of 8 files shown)


unaizorrilla:
Looks good, I try to review/merge this week!

unaizorrilla:
Hi @Daniel15 

Thanks for this contribution, a new release build is on appveyor, when finished a new packages 3.1.3 will be on NuGet with these changes!

PR #570, opened 2020-06-26
Fix memory leak, dispose MongoDb cursor

ittennull:
This PR fixes a memory leak that happens when you get a disposable cursor from MongoDb and don't dispose it.
Also I changed methods to return only names because if you can get the names then you can get the rest, and the name string is lighter than the whole BsonDocument with bunch of other details


unaizorrilla:
Thanks for contribute!

PR #623, opened 2020-09-07
[Kafka] Propagate Timeout to Kafka Health Check

KarimDarwish:

Previously, the timeout property when used with `.AddKafka()` had no impact as the `cancellationToken` was never passed, producing the message would not time out in the specified time span.


unaizorrilla:
Hi @KarimDarwish

Merged, thanks for contribute!

A new release build is on appveyor, when finished a new package with version 3.1.3 will be available on NuGet!

Issue #598, opened 2020-07-30
GetHealthReport Json parsing error


attiqeurrehman:
**What happened**:

Giving error on staging server regarding JSON parsing. Following is the error:

```json
{
    "Timestamp": "2020-07-29T15:51:06.9109870+02:00",
    "Level": "Error",
    "MessageTemplate": "GetHealthReport threw an exception when trying to get report from http://localhost:7000/health configured with name Services.",
    "Exception": "Newtonsoft.Json.JsonReaderException: Unexpected character encountered while parsing value: <. Path '', line 0, position 0.\r\n   at Newtonsoft.Json.JsonTextReader.ParseValue()\r\n   at Newtonsoft.Json.JsonReader.ReadAndMoveToContent()\r\n   at Newtonsoft.Json.JsonReader.ReadForType(JsonContract contract, Boolean hasConverter)\r\n   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)\r\n   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)\r\n   at Newtonsoft.Json.JsonSerializer.Deserialize(JsonReader reader, Type objectType)\r\n   at Newtonsoft.Json.JsonConvert.DeserializeObject(String value, Type type, JsonSerializerSettings settings)\r\n   at Newtonsoft.Json.JsonConvert.DeserializeObject[T](String value, JsonSerializerSettings settings)\r\n   at Newtonsoft.Json.JsonConvert.DeserializeObject[T](String value)\r\n   at System.Net.Http.HttpResponseMessageExtensions.As[TContent](HttpResponseMessage response)\r\n   at HealthChecks.UI.Core.HostedService.HealthCheckReportCollector.GetHealthReport(HealthCheckConfiguration configuration)",
    "Properties": {
        "SourceContext": "HealthChecks.UI.Core.HostedService.HealthCheckReportCollector",
        "Scope": [
            "HealthReportCollector is collecting health checks results."
        ],
        "ExceptionDetail": {
            "HResult": -2146233088,
            "Message": "Unexpected character encountered while parsing value: <. Path '', line 0, position 0.",
            "Source": "Newtonsoft.Json",
            "LineNumber": 0,
            "LinePosition": 0,
            "Path": "",
            "Type": "Newtonsoft.Json.JsonReaderException"
[...]

PR #627, opened 2020-09-08
Fix for health check exceptions not parsing correctly

kmcoulson:
Changed UIHealthReportEntry.Exception property to be dynamic type.

In HealthCheckReportCollector.cs (line 103) the collector tries to parse the response from a healthz endpoint to a UIHealthReport but if any of the Entries have an exception present (which will be an object) it fails to parse because the Exception property on UIHealthReportEntry is a string. 

The result is that you lose the individual entries for the different checks and the whole endpoint reports unhealthy with a parsing exception (e.g. "Unexpected character encountered while parsing value: {. Path 'entries.mongoDb.exception', line 1, position 76.").

By changing this to a dynamic the healthreport will parse correctly and the exception (as a dynamic) will be reported out correctly in the description.

#598 "Unexpected character encountered while parsing value..."

No

### DIFF
--- a/src/HealthChecks.UI.Core/UIHealthReport.cs
+++ b/src/HealthChecks.UI.Core/UIHealthReport.cs
@@ -81,7 +81,7 @@ namespace HealthChecks.UI.Core
         public IReadOnlyDictionary<string, object> Data { get; set; }
         public string Description { get; set; }
         public TimeSpan Duration { get; set; }
-        public string Exception { get; set; }
+        public dynamic Exception { get; set; }
         public UIHealthStatus Status { get; set; }
     }
 }


CarlosLanderas:
Hello @kmcoulson . Thanks for the PR!

I prefer not relying on dynamic due to it's performance. The HealthCheckReportCollector should query endpoints that speak the UI language by using the UIResponseWriter that we have in the documentation. If something is having problems with serialization is probably a configuration problem that and it's catch and throw  as an exception.

Let me dig into #598 first  please

kmcoulson:
Hi @CarlosLanderas, thanks for the reply. Yeah a dynamic isn't ideal but not sure what else would work. Even using the UIResponseWriter if the health checker returns any exception it will be an object and it looks like you then try and parse that into a string (UIHealthReportEntry.Exception) which will always fail.

I may well be missing something but if the health check returns an exception I can't see how it will ever work currently with any configuration. It looks as though one part of the code is expecting Exception to be a string (HealthCheckReportCollector.cs line 103) and another part is expecting an object (UIHealthReport.cs line 42).

kmcoulson:
FYI, #598 isn't exactly the error I was getting (see my PR comment) but is similar.

CarlosLanderas:
@kmcoulson, if the response from the healthchecks fails deserialization, the error is catch and a UnHealthy UI report is generated to display on the UI interface.

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/8ad48000ce282a517537f22fd5a0ae771a7ccbf6/src/HealthChecks.UI.Core/UIHealthReport.cs#L55

CarlosLanderas:
Let me try to reproduce your problem adding different entries to an endpoint

kmcoulson:
As an example, I used the RabbitMQ check and then killed the container running rabbit, that caused the healthcheck to include an exception.

kmcoulson:
Any joy with this?

sungam3r:
@kmcoulson Please rebase onto `master` to verify your changes by CI. Also note that deserialization was changed in #975 and uses System.Text.Json now. I tend to agree with @CarlosLanderas but know not much about repo internals.

sungam3r:
Closed as inactive.

Issue #642, opened 2020-09-23
Release 3.1.2 broke retrieval of ui-settings on relative api path


bogataj:
**What happened**: Release 3.1.2 broke retrieval of ui-settings on relative api path.

**What you expected to happen**: When reading ui-settings from api, it should respect and use the setting of api relative paths.

bogataj:
#643

CarlosLanderas:
Looks good. Thanks for catching this 👍

CarlosLanderas:
I'm gonna publish version 3.1.3 with the fix

bogataj:
> Version 3.1.3 will include this fix in some minutes

I've deployed and tested version 3.1.3 now in our kuberenetes cluster and can confirm it's working as excepted again 👍

CarlosLanderas:
@bogataj are you using the K8s operator to deploy the UI and monitor de services? :). Just to know 👍🏻

bogataj:
> @bogataj are you using the K8s operator to deploy the UI and monitor de services? :). Just to know 👍🏻

No, only plain yaml resource scripts, no CRDs, yet :)

PR #643, opened 2020-09-23
Repect the settings used on UseRelativeApiPath when retrieving ui-settings

bogataj:
Respect the settings used on UseRelativeApiPath when retrieving ui-settings

**What this PR does / why we need it**: Previous change broke the retrival of ui-settings on hosts requiring relative pathing.

**Which issue(s) this PR fixes**: #642


CarlosLanderas (commented):
Looks good! Thanks for catching this!

PR #662, opened 2020-10-27
Option to build connection string with a factory

gronbuske:
Adds an additional health check builder extension in case there is need to use an instance of a class set up using dependency injection to retrieve the connection string.

Adding additional health check builder extension using a factory to get the connection string


CarlosLanderas:
Hello @gronbuske, could you add a UnitTest to test this overload registers the hc correctly?.

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/tree/master/test/UnitTests/DependencyInjection/Npgsql

Thanks

sungam3r:
I see desired code in `master`, closing.

Issue #150, opened 2019-04-11
SQL Server Health Check with Managed Identity (Azure Access Token)


johnayoub:
The current API doesn't allow connecting to Azure SQL Server using managed identity and an access token!

Any plans to add this feature or any suggested workaround in the meantime?

unaizorrilla:
Hi, 

Interesting!! Let me check this. Of course contributions are open!! 👍

inkel:
I've similar issues with this and ended up creating a custom SQL Server health check: we use a custom connection manager, so this health check uses that, but while working on that I thought about how this feature could be added to `AspNetCore.HealthChecks.SqlServer`.

There's a tutorial named [_Secure Azure SQL Database connection from App Service using a managed identity_](https://docs.microsoft.com/en-us/azure/app-service/app-service-web-tutorial-connect-msi) that does the following once the connection is created:

```csharp
var conn = (System.Data.SqlClient.SqlConnection)Database.GetDbConnection();
conn.AccessToken = (new Microsoft.Azure.Services.AppAuthentication.AzureServiceTokenProvider()).GetAccessTokenAsync("https://database.windows.net/").Result;
```

In this case `conn` could be replaced with `conn = SqlConnection(_connectionString)` in [`src/HealthChecks.SqlServer/SqlServerHealthCheck.cs`](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/83047f04bdb1bdc8b641208dd5c36e7a0599d372/src/HealthChecks.SqlServer/SqlServerHealthCheck.cs), however that way of getting the access token is only for Azure managed SQL Databases. At my company we get the access token in a different way:

```csharp
var credential = new ClientCredential(clientId, clientSecret);
var authenticationContext = new AuthenticationContext(databaseAuthority);
var authenticationResult = await authenticationContext.AcquireTokenAsync(target, credential);
var accessToken = authenticationResult.AccessToken;
```

Given that there might be many different ways to specify the value of [`SqlConnection.AccessToken`](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlconnection.accesstoken), I think that the easiest way to implement this is to change the `SqlServerHealthCheck` constructor to either accept an optional `string accessToken = null` value, or an overloaded that accepts an `SqlConnection` instead.

[...]

PR #664, opened 2020-11-01
Add beforeOpen action to SqlServerHealthCheck

JoeSainsburys:
Adds a beforeOpen action to SqlServerHealthCheck.cs This allows consumers to do things like pass access tokens to SQL and use Managed Identities.

Please reference the issue this PR will close: #150 

Added test may not be in the correct place. The implementation of SqlServerHealthCheck does not seem to be tested, I just wanted to test it. Feel free to move/remove test.

Adds some overloads/optional parameters.

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [ ] End-to-end tests passing - they didn't pass on my machine to begin with - please can someone confirm they pass now?
- [ ] Extended the documentation - N/A? Let me know where to document it and I'll add it
- [ ] Provided sample for the feature - N/A? Let me know where to document it and I'll add it


inkel (approved):
This LGTM! Thank you @JoeSainsburys for taking the time to write this PR!

MAOliver:
@JoeSainsburys @inkel Is there something we could do to this to make it pull your token provider from the DI Container?

e.g.             

Parameter to the 2nd constructor in the healthcheckbuilderextension:

 Func<IServiceProvider, Action<SqlConnection>> beforeOpen = default):

Call to the 2nd constructor from the first constructor:

            return builder.AddSqlServer(_ => connectionString, healthQuery, name, failureStatus, tags, timeout, _ => beforeOpen);

And in the Startup.cs:

                    beforeOpen: sp => async conn => conn.AccessToken = await sp.GetRequiredService<IDBAuthTokenService>().GetTokenAsync().ConfigureAwait(false)               

I know that's a little clunky and confusing, but it seems like having it pull the tokenprovider from the service provider would be a good idea...

JoeSainsburys:
@MAOliver we certainly could - but I'm not sure if we want to (it's up to the maintainers really.) Basically what we have now is a hook for doing something to the SqlConnection, so it's very generic at the moment. If we just had an interface like IDBAuthTokenService then it is up to the developer to provide the implementation so it is still pretty generic and wouldn't require any additional nuget packages etc. I don't mind either way @inkel ?

inkel:
Sorry for the long delay. I think that the easiest, the better, at least for a first iteration. Having the before open hook gives the user enough room to execute any other kind of actions besides setting the auth token. If you open the door to add an `IDBAuthTokenService` then you also a can of worms for adding any other kind of interfaces that are injected via DI. In any case, if I were to add DI for this, I'd make it to accept an `SqlConnection` instead.

MAOliver:
Sorry, I wasn't looking for the IDBAuthTokenService -- I would just like some kind of implementation that allows you to pull some kind of token factory out of the DI container (as the connection string provider does). I think, if I'm not mistaken, that the AzureTokenProvider uses a cache that would be shared between request contexts, so, it's not a huge issue if it's implemented as is.

Thanks.

JoeSainsburys:
I was about to knock up a quick example of how I see this being used and I get what you mean now. Do you mean essentially changing the signature of beforeOpen to Action<IServiceProvider, SqlConnection> ? If so then yes I can see how that would be helpful.

JoeSainsburys:
Also please can someone address:

> End-to-end tests passing - they didn't pass on my machine to begin with - please can someone confirm they pass now?

Can I tick these two off?

>  Extended the documentation - N/A? Let me know where to document it and I'll add it
>  Provided sample for the feature - N/A? Let me know where to document it and I'll add it

And I'm not sure what your release cadence looks like but might we be able to get this in to the next release please?

unaizorrilla:
Hi @JoeSainsburys 

I',m reviewing this right now!

unaizorrilla:
Merged!

A new release build is on appveyor, when finished a new package 3.2.0 for SqlServer will be available!

JoeSainsburys:
Brilliant, thank you :smiley:

MAOliver:
@JoeSainsburys Sorry I'm late in responding. Yeah, that's kind of what I was thinking. We'll make it work either way, the internal caching for the tokenprovider makes it less necessary an issue. Thanks for contributing the fix, sir : )

Issue #650, opened 2020-10-07
Net5.0 Support


kooshan:
With release of Net5.0 RC1 we are planning to do an upgrade. UI seems incompatible with Net5.0 or EFCore5.0.

```
System.TypeLoadException: Method 'Create' in type 'Microsoft.EntityFrameworkCore.Sqlite.Query.Internal.SqliteQueryableMethodTranslatingExpressionVisitorFactory' from assembly 'Microsoft.EntityFrameworkCore.Sqlite, Version=3.1.3.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' does not have an implementation.
   at Microsoft.Extensions.DependencyInjection.SqliteServiceCollectionExtensions.AddEntityFrameworkSqlite(IServiceCollection serviceCollection)
   at Microsoft.EntityFrameworkCore.Sqlite.Infrastructure.Internal.SqliteOptionsExtension.ApplyServices(IServiceCollection services)
   at Microsoft.EntityFrameworkCore.Internal.ServiceProviderCache.ApplyServices(IDbContextOptions options, ServiceCollection services)
   at Microsoft.EntityFrameworkCore.Internal.ServiceProviderCache.<>c__DisplayClass4_0.<GetOrAdd>g__BuildServiceProvider|3()
   at Microsoft.EntityFrameworkCore.Internal.ServiceProviderCache.<>c__DisplayClass4_0.<GetOrAdd>b__2(Int64 k)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Microsoft.EntityFrameworkCore.Internal.ServiceProviderCache.GetOrAdd(IDbContextOptions options, Boolean providerRequired)
   at Microsoft.EntityFrameworkCore.DbContext..ctor(DbContextOptions options)
   at HealthChecks.UI.Core.Data.HealthChecksDb..ctor(DbContextOptions`1 options)
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
[...]

PR #678, opened 2020-11-20
upgrade to .net 5.0

maikebing:
** Net 5.0 **

Please reference the issue this PR will close: #650

### DIFF
--- a/src/HealthChecks.SignalR/SignalRHealthCheck.cs
+++ b/src/HealthChecks.SignalR/SignalRHealthCheck.cs
@@ -32,7 +32,7 @@ namespace HealthChecks.SignalR
             }
             finally
             {
-                await connection?.DisposeAsync();
+                 connection?.DisposeAsync().GetAwaiter();
             }
         }
     }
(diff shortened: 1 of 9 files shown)


CarlosLanderas:
Thanks for the PR @maikebing . We have a migration branch already to migrate to 5.0 as there are some pending features to merge in the platform so I do not thing we can integrate this in the ongoing work :/

src/HealthChecks.SignalR/SignalRHealthCheck.cs
```diff
@@ -32,7 +32,7 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
             }
             finally
             {
-                await connection?.DisposeAsync();
+                 connection?.DisposeAsync().GetAwaiter();
```
sungam3r, line 35: ?

CarlosLanderas: In my pr this is written like connection?.DisposeAsync.AsTask();

PR #681, opened 2020-11-22
Add Support For Azure Storage Options (Client Version)

GGJason:
Azure Blob Health Checks Need Support For Custom Client Version

No existing PR.
The original is that I use AzureStorage Health Check and I got an unhealty result with "Unhealty with 'null'" message.
I clone the source code and find the blob client return Header error that I solve in my application with set the client version to earlier version.
So I need to set the version of blob client to use the health check extension.
I fix it with adding an optional parameters in AddAzureBlobStorage Function and the related configurations.

Yes.

- [X] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
+++ b/src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
@@ -16,21 +16,24 @@ namespace HealthChecks.AzureStorage
 
         private readonly string _connectionString;
         private readonly string _containerName;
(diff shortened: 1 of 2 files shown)


unaizorrilla:
Hi @GGJason

Merged, on features/net50 I try to change the parameter order for BlobClientOptions. In this versions (3.3.1) this was a breaking change and I preserve your order!

A new release build is on appveyor, when finished a new package 3.3.1 will be published!

GGJason:
Thank you. It is very awesome to get such a fast response.

PR #691, opened 2020-11-27
Add an optional func parameter to SqlServer healthcheck for validatio…

CharlyAndersonUK:
…n of the returned object from the SQL query.

Currently SqlSever healthcheck executes a query which for Healthy status just needs to execute. This change adds an optional parameter to allow the result object to be validated by the user providing a func which is passed the result object.

Yes, There is a new optional parameter. If this parameter is not supplied by the user the functionality is backwards compatible.

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/src/HealthChecks.SqlServer/SqlServerHealthCheck.cs
+++ b/src/HealthChecks.SqlServer/SqlServerHealthCheck.cs
@@ -34,7 +34,7 @@ namespace HealthChecks.SqlServer
                     {
                         command.CommandText = _sql;
                         var queryResult = await command.ExecuteScalarAsync(cancellationToken);
-                        if (null != _queryValidator)
+                        if (_queryValidator != null)
                         {
                             return _queryValidator(queryResult);
                         }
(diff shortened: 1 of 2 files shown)


src/HealthChecks.SqlServer/SqlServerHealthCheck.cs
```diff
@@ -31,7 +33,11 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
                     using (var command = connection.CreateCommand())
                     {
                         command.CommandText = _sql;
-                        await command.ExecuteScalarAsync(cancellationToken);
+                        var queryResult = await command.ExecuteScalarAsync(cancellationToken);
+                        if (null != _queryValidator)
```
sungam3r, line 37:
```suggestion
                        if (_queryValidator != null)

```

sungam3r:
@carlosrecuero  What do you think? I consider that if approved 
1. such feature should be added in all database health checks that work with ADO.NET API (SqlServer, Oracle, NpgSql, etc).
2. I doubt about `ExecuteScalar` API. Why not `ExecuteReader` or `ExecuteNonQuery`? _queryValidator_ argument allows user to do something with result but now that result is fixed by only integral values. Why not to allow user to decide what API to use to get this result?
3. Previous point implies breaking changes - adding new option classes to group all parameters.

sungam3r:
And as always one more option available - keep things as is.

sungam3r:
@CharlyAndersonUK Thanks for contribution. Replaced by #1592 .

PR #738, opened 2021-03-07
Optimized RavenDB HealthChecks

ppekrol:
- do not create DocumentStore on each healthcheck - initialization performs additional and unnecessary calls to the server causing a lot of traffic
- use more lightweight endpoints to perform the healthchecks


unaizorrilla:
Merged! Thanks for contribute this!

A release build is on appveyord, when finish a new package 5.0.2 for ravendb will be available on NuGet.!

PR #754, opened 2021-03-31
Normalize IHealthChecksBuilder extensions

engenb:
Please consider this as a quick example/POC.  I wanted to get some feedback from the primary maintainers of this project before I spent time adding similar overloads across the many health check extensions.  I don't want to sink a lot of time into this if there's no desire or intention to extend the framework in this way.

There are two candidate approaches I'm proposing.

## Option 1: configure via `IOptions`
Where some health check `IHealthCheckBuilder` extensions may provide an `Action<XXXHealthCheckOptions> setup`, this first approach makes use of the `IOptions` framework by invoking `builder.Services.Configure(setup);`
### advantages:
 - `Microsoft.Extensions.Options` is already a dependency of `Microsoft.Extensions.Diagnostics.HealthChecks`
 - API doesn't change for health check extensions that have a similar `Action<XXXHealthCheckOptions>` parameter
 - developers can take advantage of `IConfigureOptions<T>` to load config however they want, with full DI support
 - by resolving `IOptionsMonitor<T>`, health check options will resolve to their current runtime values when the health check is invoked.  with the current implementation, health check options are set once at application start.
   - an example use case: I have requirements to use [Azure App Configuration](https://docs.microsoft.com/en-us/azure/azure-app-configuration/overview).  Our applications/services must react to runtime configuration changes and so must our health checks.
### disadvantages:
 - let me know if you see any, because I don't

Note: I did notice some health checks take different approaches to this already.  I saw some accepting a `Func<IServiceProvider, XXXOptions>`

## Option 2: configure directly from `IServiceProvider`
The second approach does not make use of the `IOptions<T>` framework.  Instead, the developer can provide a `Action<IServiceProvider, XXXHealthCheckOptions> setup` which is invoked when the health check is resolved from DI
### advantages:
 - does not depend on `Microsoft.Extensions.Options`
 - the intention of this api is more clear
### disadvantages:
 - not as flexible as option 1

The general goal here is to normalize how health checks are configured.  There seem to be a variety of approaches in use and deferring configuration management to the `IOptions` framework would open up a lot of flexibility for developers in a standard, familiar way.

If the developer adding health checks is the "user", then yes, this introduces non-breaking changes.  If the "user" refers to someone accessing health checks UI, then no.

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/src/HealthChecks.Network/DependencyInjection/NetworkHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.Network/DependencyInjection/NetworkHealthCheckBuilderExtensions.cs
@@ -208,7 +208,10 @@ namespace Microsoft.Extensions.DependencyInjection
             Action<SmtpHealthCheckOptions> setup, string name = default, HealthStatus? failureStatus = default,
             IEnumerable<string> tags = default, TimeSpan? timeout = default)
         {
-            builder.Services.Configure(setup);
+            if (setup != null)
+            {
+                builder.Services.Configure(setup);
+            }
 
             return builder.Add(new HealthCheckRegistration(
                 name ?? SMTP_NAME,
--- a/src/HealthChecks.Network/DependencyInjection/NetworkHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.Network/DependencyInjection/NetworkHealthCheckBuilderExtensions.cs
@@ -205,13 +205,10 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <param name="timeout">An optional System.TimeSpan representing the timeout of the check.</param>
         /// <returns>The <see cref="IHealthChecksBuilder"/>.</returns>
         public static IHealthChecksBuilder AddSmtpHealthCheck(this IHealthChecksBuilder builder,
-            Action<SmtpHealthCheckOptions> setup, string name = default, HealthStatus? failureStatus = default,
+            Action<SmtpHealthCheckOptions> setup = null, string name = default, HealthStatus? failureStatus = default,
             IEnumerable<string> tags = default, TimeSpan? timeout = default)
         {
-            if (setup != null)
-            {
-                builder.Services.Configure(setup);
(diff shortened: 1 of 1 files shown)


sungam3r:
@engenb Please rebase onto `master`. Then I can review your changes.

sungam3r:
@engenb Thanks for suggestion. I understand what you described. In general, I agree that unifying calls is good. In the case of health checks, it will be effective only for ones that have the appropriate options classes. Consider `OracleHealthCheck` for example. Second, I don't think it will be of great importance to bring `IOptions` support here. Yes, sometimes it useful, but not always. I see that the design of health checks is inclined towards the use of `Action<IServiceProvider, ...>/Func<IServiceProvider, ...>` setup delegates. Of course, technically we _can_ rewrite all health checks to migrate to `IOptions`  stuff but I doubt the value.

engenb:
The value (at least, for me) is I can load my configuration from json, azure app config service, or any other supported configuration sources, bind that config to objects, and then use those objects (with connection strings, etc) to configure health checks the same way I configure those dependencies elsewhere in my application.

Without this, I either have to hard-code configuration when setting up health checks, or "manually" pull keyed sections/values out of my `IConfiguration` which feels like an unnecessary redundancy when I'm already using `IOptions<T>` elsewhere in my application.  The underlying registration already uses `IServiceProvider` so I thought it'd be helpful to expose this to implementors of the framework who need/want to configure healthchecks from DI.

sungam3r:
Of course you _may_ add additional extension methods/overloads especially if you don't introduce breaking changes. If you have time and desire to unify all health checks across this repo, then OK, go ahead and I'll review your changes.

sungam3r:
@engenb Thanks for contribution. As I said later 
> If you have time and desire to unify all health checks across this repo, then OK, go ahead and I'll review your changes.

I close this demo draft PR for now.

PR #757, opened 2021-04-08
Connection string containing hub name

khanage:
Added overload for the AddAzureEventHub method to infer whether to add
the EntityPath to the connection string

In the event that the Azure Event Hub connection string includes an event hub name, the provided event hub name is ignored. This is a little awkward however, as one must provide a dummy name, or duplicate the name, parse the name, etc.

This PR resolves the problem by allowing a full connection string to be passed in.

Happy to take any suggestions :)

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
+++ b/src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
@@ -28,7 +28,7 @@ namespace HealthChecks.AzureServiceBus
             if (!connectionString.Contains(EntityPathSegment))
             {
                 throw new ArgumentException(
-                    "Connection string did not contain an event hub name.",
+                    "Connection string should contain event hub name.",
                     nameof(connectionString)
                 );
             }
--- a/src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
+++ b/src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
@@ -13,8 +13,28 @@ namespace HealthChecks.AzureServiceBus
     {
         const string EntityPathSegment = "EntityPath=";
 
-        private static readonly ConcurrentDictionary<string, EventHubProducerClient> _eventHubConnections = new ConcurrentDictionary<string, EventHubProducerClient>();
+        private static readonly ConcurrentDictionary<string, EventHubProducerClient> _eventHubConnections =
+            new ConcurrentDictionary<string, EventHubProducerClient>();
+
         private readonly string _eventHubConnectionString;
+
+        public AzureEventHubHealthCheck(string connectionString)
+        {
+            if (string.IsNullOrEmpty(connectionString))
+            {
+                throw new ArgumentNullException(nameof(connectionString));
+            }
+
+            if (!connectionString.Contains(EntityPathSegment))
+            {
+                throw new ArgumentException(
+                    "Connection string did not contain an event hub name.",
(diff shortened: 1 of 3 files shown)


src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
```diff
@@ -13,50 +13,78 @@ public class AzureEventHubHealthCheck
...
+        {
+            if (string.IsNullOrEmpty(connectionString))
+            {
+                throw new ArgumentNullException(nameof(connectionString));
+            }
+
+            if (!connectionString.Contains(EntityPathSegment))
+            {
+                throw new ArgumentException(
+                    "Connection string did not contain an event hub name.",
```
sungam3r, line 31:
```suggestion
                    "Connection string should contain event hub name.",
```

src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
```diff
@@ -13,50 +13,78 @@ public class AzureEventHubHealthCheck
...
-                    }
+                if (!_eventHubConnections.TryGetValue(_eventHubConnectionString, out var producer))
+                {
+                    producer = new EventHubProducerClient(_eventHubConnectionString);
+
+                    if (!_eventHubConnections.TryAdd(_eventHubConnectionString, producer))
+                    {
+                        return new HealthCheckResult(context.Registration.FailureStatus,
+                            description: "EventHubProducerClient can't be added into dictionary.");
+                    }
```
sungam3r, line 87: Your PR contains a lot of not relevant changes. For example, what exactly changed in this code block?

src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
```diff
@@ -68,4 +96,4 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
             }
         }
     }
-}
+}
```
sungam3r, line 99: This change is not needed.

unaizorrilla:
Hi @khanage 

Thanks for send this PR! My concern here is that new overload is a breaking change because can break some constructions using the name of the healthcheck as the name of the eventhub and introduce more confussion. Please let me know if you have more comments on this!


Thanks!

src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
khanage: It's just whitespace from autoformatting apologies. I can revert these changes if you'd prefer?

khanage:
@unaizorrilla do you mean that name might be selected for existing constructors instead of event hub name? The order is the same so it shouldn't break - happy to make some changes to this though if you have suggestions?

src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
sungam3r: It does not matter to me, ask the maintainers.

sungam3r:
@khanage Please rebase onto `master` if you are going to continue work on this PR so I can review/merge.

khanage:
Apologies for leaving this stale, it looks like this has already landed in main so I'll abandon.

Issue #651, opened 2020-10-08
HealthCheck collector HostedService threw an error: Value cannot be null. (Parameter 'source')


AlexanderArvanitis:
What happened:
Health checks UI is empty although the /health endpoint returns an response. When running on Kestrel I can tell from the logs that the the endpoint is called and returns correctly.

From the exception message I gather that execution.Entries in SaveExecutionHistoryEntries might be empty for some reason, but I haven't been able to test it.

Maybe I'm doing something wrong with how I setup it all up?
**This is my Startup.cs:**
```
public void ConfigureServices(IServiceCollection services)
        {
            services
                .AddHealthChecksUI(setup => 
                {
                    setup.AddHealthCheckEndpoint("test", "https://localhost:8080/health");
                })
                .AddInMemoryStorage();
        }
public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
        {
            if (env.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }

            app.UseRouting();

            app.UseEndpoints(endpoints =>
            {
                endpoints.MapGet("/", async context =>
                {
                    await context.Response.WriteAsync("Hello World!");
                });
                endpoints.MapHealthChecksUI();
            });
        }

```
**From the outputwindow/kestrell:**
```
info: HealthChecks.UI.Core.HostedService.UIInitializationHostedService[0]
      Initializing UI Database
info: HealthChecks.UI.Core.HostedService.UIInitializationHostedService[0]
      Saving healthchecks configuration to database
info: Microsoft.Hosting.Lifetime[0]
      Now listening on: https://localhost:5001
info: Microsoft.Hosting.Lifetime[0]
      Now listening on: http://localhost:5000
info: System.Net.Http.HttpClient.health-checks.LogicalHandler[100]
      Start processing HTTP request GET https://localhost:8080/health
info: System.Net.Http.HttpClient.health-checks.ClientHandler[100]
      Sending HTTP request GET https://localhost:8080/health
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
info: Microsoft.Hosting.Lifetime[0]
      Hosting environment: Development
info: Microsoft.Hosting.Lifetime[0]
      Content root path: C:\Users\000-alar\source\repos\HealthMonitor\HealthMonitor
info: System.Net.Http.HttpClient.health-checks.ClientHandler[101]
[...]

PR #794, opened 2021-06-30
Issue 651 - Fixed an issue in UI when deserializing invalid UIHealthReport

liquidsnk:
#651 occurs when a health endpoint returns JSON that deserializes to a UIHealthReport but not necessarily a vaild one.
For example in my case, the health endpoint was returning a 404 problem details which deserialized to a UIHealthReport with 404 status (instead of healthy/unhealthy) and a null Entries dictionary (which ended triggering the exception).
This PR addresses the validations in a JsonConverter added and passed to the deserialization.

Please reference the issue this PR will close: #651

Something even simpler would have been to just add the validations inline after the As<UIHealthReport>() call but I felt tying them to the deserialization itself would be better, with some more work maybe it could be made the default converter for the class.

No

- [X] Code compiles correctly
- [ ] Created/updated tests
- [X] Unit tests passing

### DIFF
--- a/src/HealthChecks.UI/Core/Extensions/HttpResponseMessageExtensions.cs
+++ b/src/HealthChecks.UI/Core/Extensions/HttpResponseMessageExtensions.cs
@@ -5,7 +5,7 @@ namespace System.Net.Http
 {
     public static class HttpResponseMessageExtensions
     {
-        public static async Task<TContent> As<TContent>(this HttpResponseMessage response)
+        public static async Task<TContent> As<TContent>(this HttpResponseMessage response, params JsonConverter[] converters)
         {
             if (response != null)
             {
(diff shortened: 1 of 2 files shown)


sungam3r:
@liquidsnk Please rebase onto `master` if you are going to continue work on this PR. Also note that #975 changes serializer to System.Text.Json. If you still experiencing this issue then please add test to verify your changes.

sungam3r:
Ping @liquidsnk

liquidsnk:
Hey @sungam3r , sorry. This is probably still happening, but I haven't looked in a while. I'll try it out soon.

liquidsnk:
Seems it not an issue anymore!, at least not for my case. The new deserialization runs the constructor of the class and that takes care of the problem. I'll close my PR.

Issue #783, opened 2021-06-03
BlobService check requires Storage Account Owner permissions


richardjharding:
Make the call to Get Blob Service Properties optional - perhaps based on the Options object used in this check

As currently implemented this line https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/a5084763d87001033704c219696cdca43dfcfede/src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs#L44

Seems to require a high level of authorisation - according to the ms docs it needs storage account owner role - https://docs.microsoft.com/en-us/rest/api/storageservices/get-blob-service-properties#authorization

In my case I was looking to limit the necessary permissions for my app to just those required to access the blob container but adding this health check means I need to grant excessive permissions (imho :))

Meertman:
We had the same problem when using the health checks in an enterprise context. The account used by our application does not have the storage account owner permission, but we would like to perform the health check.

MrKevHunter:
Any idea when this will be merged, the PR looks good?

PR #806, opened 2021-07-19
Allow AzureBlobHealthCheck to work with "Storage Blob Data Contributor"

Nisden:

This pull request allows for the AzureBlobHealthCheck to use Azure AD roles for authentication.
Specially it allows an user with the role `Storage Blob Data Contributor` to run the health check.

This will fix the issue #783

`Storage Blob Data Contributor` is an Azure AD RBAC role that can read and write data to storage containers. Its typically used together with Managed Identities. The `Storage Blob Data Contributor` only have a very limited set of permissions (See image)

image

The GetServiceProperties requires account ownership, that you generally don't wanna delegate to your account.
As a note, an Storage Account Key equals full account ownership.

There should be no noticeable changes.

- [x] Code compiles correctly
- [ ] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
+++ b/src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
@@ -41,9 +41,9 @@ namespace HealthChecks.AzureStorage
             try
             {
                 var blobServiceClient = GetBlobServiceClient();
-                await foreach (var container in blobServiceClient.GetBlobContainersAsync(cancellationToken: cancellationToken))
+                await foreach (var page in blobServiceClient.GetBlobContainersAsync(cancellationToken: cancellationToken).AsPages(pageSizeHint: 1))
                 {
-                    // We are just validating that we have access to list the containers.
+                    break;
                 }
 
                 if (!string.IsNullOrEmpty(_containerName))
--- a/src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
+++ b/src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
@@ -41,7 +41,10 @@ namespace HealthChecks.AzureStorage
             try
             {
                 var blobServiceClient = GetBlobServiceClient();
-                var serviceProperties = await blobServiceClient.GetPropertiesAsync(cancellationToken);
+                await foreach (var container in blobServiceClient.GetBlobContainersAsync(cancellationToken: cancellationToken))
+                {
+                    // We are just validating that we have access to list the containers.
+                }
 
                 if (!string.IsNullOrEmpty(_containerName))
                 {


Meertman (commented):
Is it necessary to loop over all containers? Would it not be just enough to see that the operation returns?

src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
```diff
@@ -41,7 +41,10 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
             try
             {
                 var blobServiceClient = GetBlobServiceClient();
-                var serviceProperties = await blobServiceClient.GetPropertiesAsync(cancellationToken);
+                await foreach (var container in blobServiceClient.GetBlobContainersAsync(cancellationToken: cancellationToken))
```
Meertman, line 44: Is it necessary to loop over all containers? Would it not be just enough to see that the operation returns?

Meertman (changes requested):

src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
Nisden: I will have to check up on that, as I don't know the exact behaviour of an IAsyncEnumerable.
But I imagine that we might be able to just read the first element.

MrKevHunter:
Can this get merged? we really need it

Nisden:
Any chance you can review and merge this pull request @sungam3r its been pending for quite a while, and is a minor change.

sungam3r:
I'm OK to review/merge this one when you rebase your changes onto actual master. The reason is to run CI tests that were added some time ago into master.

Nisden:
@sungam3r its been rebased, but you need to run/approve the workflow.

sungam3r:
@CarlosLanderas @unaizorrilla `global.json` looks like this
`
{
  "projects": ["src", "tests", "samples"],
  "sdk": {
    "version": "6.0.100"
  }
}
`

Now all CI will fail because they use 6.0.x version and current version is 6.0.200. Options:
1. Configure `global.json` - https://docs.microsoft.com/en-us/dotnet/core/tools/global-json?tabs=netcore3x#rollforward
2. Remove `global.json`

Honestly, I would prefer the second, I have never seen a lot of benefit in it.

unaizorrilla:
Hi @sungam3r 

+ @carlosrecuero 

What about add [rollForward](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json) to global.json? 

```json
{
  "sdk": {
    "version": "3.1.102",
    "rollForward": "latestPatch"
  }
}
```

sungam3r:
@dependabot rebase

PR #825, opened 2021-08-31
Added support for non-default http version and version policy

mattcox314:
Change enables requests to endpoints requiring http/2.

Current behaviour is that even when using `configureClient: (provider, client) => client.DefaultRequestVersion = HttpVersion.Version20` in the `AddUrlGroup` extension, the HttpRequestMessage will be created with a default Version of HttpVersion.Version11 thereby making it functionally impossible to test an endpoint that requires http/2.

This change is functionally equivalent to how the HttpRequestMessage is handled inside the HttpClient. See: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpClient.cs#L821

No

### DIFF
--- a/src/HealthChecks.Uris/UriHealthCheck.cs
+++ b/src/HealthChecks.Uris/UriHealthCheck.cs
@@ -40,6 +40,11 @@ namespace HealthChecks.Uris
 
                     var requestMessage = new HttpRequestMessage(method, item.Uri);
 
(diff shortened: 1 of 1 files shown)


sungam3r:
@mattcox314 Could you please add some tests to verify changes?

sungam3r:
OK, I checked docs and it seems to me as good to add.

Issue #788, opened 2021-06-09
Comment peut on gérer différent service avec le même nom, mais dans des espaces de nom différent


sboule745:
**What would you like to be added**: We would like to be able to filter the services by name and namespace.

**Why is this needed**: our structure is separated by namespace and this can cause problems if we find the same service name twice but in two different namespace

Issue #809, opened 2021-07-22
Multi-menu Healthchecks UI


rahul230691:
**What would you like to be added**:

I need help to create multiple menus where I could display my individual application health checks. 

**Why is this needed**:

I have 8-10 applications and each application has several microservices and has more than several health checks based on respected services. I need to create single menu for every individual application in UI. so that all health checks related to that single application could be displayed in single view. and same for others as well.

I have attached a layout for reference. Hope it helps to understand.

Thank you

![multi-menu healthchecks UI](https://user-images.githubusercontent.com/20270535/126677711-3d1b16b3-2bdc-4961-99bc-fd9bf6fbf6b0.png)

Issue #690, opened 2020-11-26
Order by Name on Health Check Status page


deathchurch:
The ability to order the health checks on the Health Check Status page by Name rather than the Id in the database.

I have added the checks in alphabetical order in appsettings.json and would like to keep that order when displaying the Health Checks Status page. I assume they get added into the DB in the order they complete for the first time.

I assume its this line https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/ad005897f467b97b6eff86848de8a60734d9bd1b/src/HealthChecks.UI/Middleware/UIApiEndpointMiddleware.cs#L46 a nice enhancement would be the ability to change the display order.

PR #826, opened 2021-08-31
Features/UI grouping

Meertman:
This PR allows the end-user to group his health check endpoints into either custom defined groups, or in case the end-user is using the kubernetes discovery it will group the health checks using the namespace in which the pods that execute the healthchecks are hosted.

**Which issue(s) this PR fixes**: #788, #690, groundwork for #809 

Please reference the issue this PR will close: #788 #690 

- [X] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


Meertman:
Credits where credits are due, the UI work has been done by @GeniaT

BoSoeborgPetersen:
I love the idea of adding groupings to the HealthCheckUI, but I had to add "dom.iterable" to tsconfig.json -> "compilerOptions" -> "lib" in HealthChecksUI project.

And still when I run the HealthChecks.UI.Branding sample I see this:
image
And this when expanded (which can only be done with the expand all button:
image

I would assume that groups are optional and that all health checks without a group would end up in a default group?

Meertman:
@BoSoeborgPetersen, sorry, I did indeed forget to check in that change to the tsconfig.json file. (I added it to the PR)

So yes, health checks without a group end up in the default group. Which currently has no name :)

BoSoeborgPetersen:
@Meertman, from what I can tell, the backend works as expected, but there are some frontend issues:

There is a problem with the expand/collapse button at the group level:
 - Default group: it does nothing
 - Other groups: it works, but it also expands the health checks in the group
I think the problem is that the typescript does not react well to the default group having null as the name
 
I would also suggest that:
 - Groups have a different background color than health checks
 - If only the default group exists, then just display the health checks without the group level (no API changes, just UI), this would ensure that current users don't see any changes because of this feature, in the UI, and they can then choose to use this feature. But maybe this should also be part of the API, so the data structure is the same for people not using this feature, otherwise we could break peoples use of the API from somewhere other than the UI.

I also found a possible mistake on line 58 in IEndpointRouteBuilderExtensions.cs, it should probably be AddAsync, instead of RemoveAsync.

I have added the background color for groups locally, and will look further at the other parts, but I am limited by my javascript/typescript/frontend skills, and I am also not sure where to push my changes, possibly a fork of your fork.

build/docker-images/HealthChecks.UI.Image/Extensions/IEndpointRouteBuilderExtensions.cs
```diff
@@ -54,15 +55,15 @@ public static class IEndpointRouteBuilderExtensions
 
                         if (type == PushServiceKeys.ServiceAdded)
                         {
-                            await pushService.AddAsync(name, uri);
+                            await pushService.RemoveAsync(name, group);
```
smbecker, line 58: Shouldn't this be `AddAsync`?

smbecker:
It would be nice to be able to include the group name in [webhook notifications](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/50594f0643515e83f9bf6ce864f9594bf980b865/src/HealthChecks.UI/Core/Notifications/WebHookFailureNotifier.cs#L78).

Meertman:
I am also limited by my knowledge of React, so if someone could take a look at the UI part, this might be merged as well?

Meertman:
> @Meertman, from what I can tell, the backend works as expected, but there are some frontend issues:
> 
> There is a problem with the expand/collapse button at the group level:
> 
> * Default group: it does nothing
> * Other groups: it works, but it also expands the health checks in the group
>   I think the problem is that the typescript does not react well to the default group having null as the name
> 
> I would also suggest that:
> 
> * Groups have a different background color than health checks
> * If only the default group exists, then just display the health checks without the group level (no API changes, just UI), this would ensure that current users don't see any changes because of this feature, in the UI, and they can then choose to use this feature. But maybe this should also be part of the API, so the data structure is the same for people not using this feature, otherwise we could break peoples use of the API from somewhere other than the UI.
> 
> I also found a possible mistake on line 58 in IEndpointRouteBuilderExtensions.cs, it should probably be AddAsync, instead of RemoveAsync.
> 
> I have added the background color for groups locally, and will look further at the other parts, but I am limited by my javascript/typescript/frontend skills, and I am also not sure where to push my changes, possibly a fork of your fork.

@BoSoeborgPetersen for your information, @Tom-De-Backer has implemented these changes and I've added them to the pull request.

@sungam3r, can we get these changes merged with the main repo?

sungam3r:
@Meertman Sorry but I'm not familiar and I don't use such areas of this repo as docker/JS/UI so better to ask maintainers. I just help to keep things moving in HC packages, CI and around some other stuff.

TrueBurn:
Is it possible to have the group come from an annotation on the service (the annotation should be configurable) then maybe with the k8s namespace as a fallback if there is no group set via the service annotation?

PR #847, opened 2021-10-29
Fix for race condition in AzureServiceBusSubscriptionHealthCheck

### LINKED ISSUE #816: HealthCheck UI on a horizontally scaled environment triggers multiple checks to non-tread-safe code
**What happened**:
We discovered some error logs in our environment:

> Health check Subscriber/xxx/xxx completed after 8.3213ms with status Unhealthy and 'New management client connection can't be added into dictionary.'

No Error Logs at all

* Run a service with an `AzureServiceBusSubscriptionHealthCheck`
* Set up HealtCheck UI in an App that is horizontally scaled to 3 instances and connect it to the previously mentioned service
* Run for some time and analyze your error logs

**Description and Analysis**:
We saw the error in a service that has an `AzureServiceBusSubscriptionHealthCheck`. The code inside this check is not Thread-Safe (access to `AzureServiceBusSubscriptionHealthCheck.ManagementClientConnections`) and can cause unhealthy results even everything is working fine. In my opinion, we should 1.) find a solution to prevent multiple health checks to the same health endpoint inside the HealthChecks UI and 2.) fix those health checks so that there is no race condition anymore.

Screenshot from our login infrastructure with 3 requests all within the same second:
Request Logs

Screenshot of code which causes the race condition:
image

**Environment**:

- .NET Core version: 3.1
- Healthchecks version:
  - `AspNetCore.HealthChecks.AzureServiceBus`: 3.2.2
  - `AspNetCore.HealthChecks.UI`: 5.0.1
- Operative system: Linux / Docker
- Others: Azure

sungam3r:
Fixes #816

### DIFF
--- a/src/HealthChecks.AzureServiceBus/AzureServiceBusSubscriptionHealthCheck.cs
+++ b/src/HealthChecks.AzureServiceBus/AzureServiceBusSubscriptionHealthCheck.cs
@@ -7,10 +7,11 @@ namespace HealthChecks.AzureServiceBus
 {
     using Azure.Core;
 
-    public class AzureServiceBusSubscriptionHealthCheck   : AzureServiceBusHealthCheck, IHealthCheck
+    public class AzureServiceBusSubscriptionHealthCheck : AzureServiceBusHealthCheck, IHealthCheck
     {
         private readonly string _topicName;
         private readonly string _subscriptionName;
+        private string _connectionKey;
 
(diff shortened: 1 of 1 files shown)


unaizorrilla:
Hi @carlosrecuero can you check this?

sungam3r:
Should I rebase to `netcore-6.0` ?

carlosrecuero (approved):

carlosrecuero:
Bug fixing was merged into master and applied in netcore6.0 branch.

PR #851, opened 2021-11-04
Unify projects: remove unnecessary properties, common naming, indentation and property order

sungam3r:
Cosmetic changes in csprojs. No user-facing changes. No actual source code changes, only configs. Note that two projects have issues with naming - `HealthChecks.IdSvr` and `HealthChecks.RabbitMQ`. For those `PackageId` != `ProjectName`. I preserved the package names/namespaces and changed only the project/folders names to match conventions from dependencies.props.

`Version` property can also be unified if you want.


build/dependencies.props
```diff
@@ -1,12 +1,13 @@
...
-  <PropertyGroup Label="SDK Versions">    
+  <PropertyGroup Label="SDK Versions">
     <NetFrameworkVersion>net5.0</NetFrameworkVersion>
     <NetStandardVersion>netstandard2.0</NetStandardVersion>
     <MicrosoftNETTestSdk>16.8.0</MicrosoftNETTestSdk>
     <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
   <PropertyGroup Label="Package information">
+    <PackageId>AspNetCore.$(MSBuildProjectName)</PackageId>
```
sungam3r, line 10: @unaizorrilla This will not copy the configuration among dozens of projects. Alas, for two of them the authors did not comply with the naming conventions and gave different names to their packages.

sungam3r:
Well... Do I understand correctly that CI is configured only in .NET 6 branch?

sungam3r:
@unaizorrilla I change base branch to `netcore-6.0` as you suggested. I have managed most of the merge problems but I still see test errors. Do I understand correctly that they were before my changes also? I had to rename `SolR` to `Solr` to `dotnet restore` worked. Current errors are:
```
System.OperationCanceledException : The operation was canceled.
Response status code does not indicate success: 503 (Service Unavailable).
Expected the enum to be HttpStatusCode.OK {value: 200}, but found HttpStatusCode.ServiceUnavailable {value: 503}
```
Note that I did not change any source file. All files are `.csproj` or `.yaml`. 4 `.cs` files were just renamed without actual changes. You can make sure this hiding whitespace changes:
изображение

.github/workflows/healthchecks_solr_ci.yml
```diff
@@ -4,15 +4,15 @@ on:
...
+      - src/HealthChecks.Solr/**
       - test/HealthChecks.SolR.Tests/**
       - .github/workflows/healthchecks_solr_ci.yml
     tags-ignore:
       - preview-*
   pull_request:
     #branches: [ master ]
     paths:
-    - src/HealthChecks.SolR/**
+    - src/HealthChecks.Solr/**
```
sungam3r, line 15: Why so ?

carlosrecuero: Sorry, i don't understand your question.

sungam3r: I had to change `R` with `r` to pass CI build. On Windows, it does not matter, the build is successful, but not on Linux. I just became interested, whether this error remained unnoticed. Or did you see this error when working with `netcore-6.0` branch?

PR #876, opened 2021-12-17
Most recent history should be on TOP.

VKAlwaysWin:
Currently API is returning history in Desc Order (which is correct), but on UI we are showing older on TOP which is not really obvious.

**Does this PR introduce a user-facing change?**: YES

It will show histories, as it is shown on documentation: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks#health-status-history-timeline


alvarodl (approved):

sungam3r:
@carlosrecuero Looks like trivial change. I don't use UI parts of this repo so I do not have sufficient knowledge to merge.

carlosrecuero:
Yes, @sungam3r, I checked it with @alvarodl last week. Thank you both.

PR #879, opened 2021-12-18
IdSvr: Add prefix to request URI

ngruson:
I have several .NET 5 APIs in which I want to check if Identity Server is already reachable. 
My Identity Server is behind an NGINX reverse proxy. The URI looks something like: https://dev.k8s-local.my-app.com/identityserver.
Both the APIs and Identity Server are part of the same Kubernetes deployment.

In described situation, the `GetAsync` call to .well-known/openid-configuration fails with the current version of the IdSvr health check.
With this fix, I can provide a prefix so that the request URI becomes identityserver/.well-known/openid-configuration.

The new prefix parameter of `AddIdentityServer` is optional.
If you don't use the prefix, code should still compile.

I tried to add unit tests, but my local build of HealthChecks.UI fails.
I just installed NPM version 8.1.4.
```
*** Bundling client files ***
1>
1>> healthchecks.ui@1.0.0 build
1>> npm run build:dll && npm run build:prod
1>
1>
1>> healthchecks.ui@1.0.0 build:dll
1>> webpack-cli --mode production --config ./build/webpack.dll.js
1>
1>node:internal/crypto/hash:67
1>  this[kHandle] = new _Hash(algorithm, xofLen);
1>                  ^
1>
1>Error : error : 0308010C:digital envelope routines::unsupported
1>    at new Hash (node:internal/crypto/hash:67:19)
1>    at Object.createHash (node:crypto:130:10)
1>    at module.exports (C:\Repos\ngruson\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\node_modules\webpack\lib\util\createHash.js:135:53)
1>    at NormalModule._initBuildHash (C:\Repos\ngruson\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\node_modules\webpack\lib\NormalModule.js:417:16)
1>    at handleParseError (C:\Repos\ngruson\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\node_modules\webpack\lib\NormalModule.js:471:10)
1>    at C:\Repos\ngruson\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\node_modules\webpack\lib\NormalModule.js:503:5
1>    at C:\Repos\ngruson\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\node_modules\webpack\lib\NormalModule.js:358:12
1>    at C:\Repos\ngruson\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\node_modules\loader-runner\lib\LoaderRunner.js:373:3
1>    at iterateNormalLoaders (C:\Repos\ngruson\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\node_modules\loader-runner\lib\LoaderRunner.js:214:10)
1>    at Array.<anonymous> (C:\Repos\ngruson\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\node_modules\loader-runner\lib\LoaderRunner.js:205:4)
1>    at Storage.finished (C:\Repos\ngruson\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\node_modules\enhanced-resolve\lib\CachedInputFileSystem.js:55:16)
1>    at C:\Repos\ngruson\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\node_modules\enhanced-resolve\lib\CachedInputFileSystem.js:91:9
1>    at C:\Repos\ngruson\AspNetCore.Diagnostics.HealthChecks\src\HealthChecks.UI\node_modules\graceful-fs\graceful-fs.js:123:16
1>    at FSReqCallback.readFileAfterClose [as oncomplete] (node:internal/fs/read_file_context:68:3) {
1>  opensslErrorStack: [ 'error:03000086:digital envelope routines::initialization error' ],
1>  library: 'digital envelope routines',
1>  reason: 'unsupported',
1>  code: 'ERR_OSSL_EVP_UNSUPPORTED'
```

### DIFF
--- a/src/HealthChecks.OpenIdConnectServer/DependencyInjection/IdSvrHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.OpenIdConnectServer/DependencyInjection/IdSvrHealthCheckBuilderExtensions.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <param name="tags">A list of tags that can be used to filter sets of health checks. Optional.</param>
         /// <param name="timeout">An optional System.TimeSpan representing the timeout of the check.</param>
         /// <returns>The <see cref="IHealthChecksBuilder"/>.</returns>
-        public static IHealthChecksBuilder AddIdentityServer(this IHealthChecksBuilder builder, Uri idSvrUri, string name = default, HealthStatus? failureStatus = default, IEnumerable<string> tags = default, TimeSpan? timeout = default)
+        public static IHealthChecksBuilder AddIdentityServer(this IHealthChecksBuilder builder, Uri idSvrUri, string name = default, HealthStatus? failureStatus = default, IEnumerable<string> tags = default, TimeSpan? timeout = default, string prefix = default)
         {
             var registrationName = name ?? NAME;
 
@@ -30,7 +30,10 @@ namespace Microsoft.Extensions.DependencyInjection
 
             return builder.Add(new HealthCheckRegistration(
                 registrationName,
-                sp => new IdSvrHealthCheck(() => sp.GetRequiredService<IHttpClientFactory>().CreateClient(registrationName)),
+                sp => new IdSvrHealthCheck(
+                    () => sp.GetRequiredService<IHttpClientFactory>().CreateClient(registrationName),
+                    prefix
+                ),
                 failureStatus,
                 tags,
                 timeout));
--- a/src/HealthChecks.OpenIdConnectServer/DependencyInjection/IdSvrHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.OpenIdConnectServer/DependencyInjection/IdSvrHealthCheckBuilderExtensions.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <param name="tags">A list of tags that can be used to filter sets of health checks. Optional.</param>
         /// <param name="timeout">An optional System.TimeSpan representing the timeout of the check.</param>
         /// <returns>The <see cref="IHealthChecksBuilder"/>.</returns>
-        public static IHealthChecksBuilder AddIdentityServer(this IHealthChecksBuilder builder, Uri idSvrUri, string name = default, HealthStatus? failureStatus = default, IEnumerable<string> tags = default, TimeSpan? timeout = default, string prefix = default)
+        public static IHealthChecksBuilder AddIdentityServer(this IHealthChecksBuilder builder, Uri idSvrUri, string name = default, HealthStatus? failureStatus = default, IEnumerable<string> tags = default, TimeSpan? timeout = default, string requestUri = default)
         {
             var registrationName = name ?? NAME;
 
@@ -32,7 +32,7 @@ namespace Microsoft.Extensions.DependencyInjection
                 registrationName,
                 sp => new IdSvrHealthCheck(
                     () => sp.GetRequiredService<IHttpClientFactory>().CreateClient(registrationName),
-                    prefix
+                    requestUri
                 ),
                 failureStatus,
                 tags,
--- a/src/HealthChecks.OpenIdConnectServer/IdSvrHealthCheck.cs
+++ b/src/HealthChecks.OpenIdConnectServer/IdSvrHealthCheck.cs
@@ -12,16 +12,22 @@ namespace HealthChecks.IdSvr
         const string IDSVR_DISCOVER_CONFIGURATION_SEGMENT = ".well-known/openid-configuration";
 
         private readonly Func<HttpClient> _httpClientFactory;
-        public IdSvrHealthCheck(Func<HttpClient> httpClientFactory)
+        private readonly string _prefix;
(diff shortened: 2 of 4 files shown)


src/HealthChecks.OpenIdConnectServer/DependencyInjection/IdSvrHealthCheckBuilderExtensions.cs
```diff
@@ -22,18 +22,21 @@ public static class IdSvrHealthCheckBuilderExtensions
...
+                    () => sp.GetRequiredService<IHttpClientFactory>().CreateClient(registrationName),
+                    prefix
+                ),
                 failureStatus,
                 tags,
                 timeout));
         }
     }
-}
+}
```
sungam3r, line 42:
```suggestion
}

```

src/HealthChecks.OpenIdConnectServer/IdSvrHealthCheck.cs
```diff
@@ -12,16 +12,22 @@ public class IdSvrHealthCheck
...
         }
         public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
         {
             try
             {
                 var httpClient = _httpClientFactory();
-                var response = await httpClient.GetAsync(IDSVR_DISCOVER_CONFIGURATION_SEGMENT, cancellationToken);
+                var requestUri = string.IsNullOrEmpty(_prefix)
+                    ? IDSVR_DISCOVER_CONFIGURATION_SEGMENT
+                    : $"{_prefix}/{IDSVR_DISCOVER_CONFIGURATION_SEGMENT}";
```
sungam3r, line 28: Maybe just move `IDSVR_DISCOVER_CONFIGURATION_SEGMENT` from constant into new parameter and use as default value?

sungam3r: Also note that the same configuration may be done via provided `httpClientFactory` so this PR not needed.

sungam3r: `AddHttpClient` calls are cumulative so you can allready configure your http client to use any prefix you want.

sungam3r:
I fexed an error with invalid path in CI. You may want to rebase.

src/HealthChecks.OpenIdConnectServer/IdSvrHealthCheck.cs
ngruson: I changed the prefix parameter into the requestUri parameter, but anyone that overrides it should also include .well-known/openid-configuration in the requestUri value now.
Also: how would this work without this PR? It's not enough to set the BaseAddress of the HttpClient. Can I set the request uri of `GetAsync` in `httpClientFactory`?

src/HealthChecks.OpenIdConnectServer/DependencyInjection/IdSvrHealthCheckBuilderExtensions.cs
```diff
@@ -22,18 +22,21 @@ public static class IdSvrHealthCheckBuilderExtensions
         /// <param name="tags">A list of tags that can be used to filter sets of health checks. Optional.</param>
         /// <param name="timeout">An optional System.TimeSpan representing the timeout of the check.</param>
         /// <returns>The <see cref="IHealthChecksBuilder"/>.</returns>
-        public static IHealthChecksBuilder AddIdentityServer(this IHealthChecksBuilder builder, Uri idSvrUri, string name = default, HealthStatus? failureStatus = default, IEnumerable<string> tags = default, TimeSpan? timeout = default)
+        public static IHealthChecksBuilder AddIdentityServer(this IHealthChecksBuilder builder, Uri idSvrUri, string name = default, HealthStatus? failureStatus = default, IEnumerable<string> tags = default, TimeSpan? timeout = default, string requestUri = default)
```
sungam3r, line 25:
```suggestion
        public static IHealthChecksBuilder AddIdentityServer(this IHealthChecksBuilder builder, Uri idSvrUri, string name = default, HealthStatus? failureStatus = default, IEnumerable<string> tags = default, TimeSpan? timeout = default, string requestUri = ".well-known/openid-configuration")
```

src/HealthChecks.OpenIdConnectServer/IdSvrHealthCheck.cs
sungam3r: If this works for your case then I really think this PR not needed.

ngruson: Before I made the PR, I did a test in a console app and thought that it didn't work with only setting BaseAddress.
It turns out that I had to end the BaseAddress with a slash which I didn't think of initially.
So when I set the base address to `https://dev.k8s-local.my-app.com/identityserver**/**', it actually works.
Looks like you're right that this PR is not needed. Shall I close it?

sungam3r: Yep.

Issue #903, opened 2022-02-09
AzureServiceBus health checks can fail due to a race condition


davecluderay:
The `AzureServiceBusQueueHealthCheck` and `AzureServiceBusTopicHealthCheck` checks can fail when running concurrently, incorrectly reporting the service as unhealthy.

The same issue was previously reported and fixed for the `AzureServiceBusSubscriptionHealthCheck` check (Bug #816, PR #847).
The same fix can be applied to [AzureServiceBusQueueHealthCheck](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.AzureServiceBus/AzureServiceBusQueueHealthCheck.cs#L38-L46) and [AzureServiceBusTopicHealthCheck](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.AzureServiceBus/AzureServiceBusTopicHealthCheck.cs#L39-L48).

Please note that `AzureServiceBusTopicHealthCheck` actually has an extra line of code that appends `_topicName` again, even though it's already there (this looks like a mistake, so should be replaced with the same thing the other two checks do).

I'm happy to do a PR for it.

sungam3r:
Fixed in #905.

PR #905, opened 2022-02-09
Fix similar race conditions in AzureServiceBusQueueHealthCheck and AzureServiceBusTopicHealthCheck.

davecluderay:
This fixes issue #903 by replacing separate calls to `TryGetValue` and `TryAdd` with a single `GetOrAdd` operation against the concurrent dictionary of client connections.

### DIFF
--- a/src/HealthChecks.AzureServiceBus/AzureServiceBusTopicHealthCheck.cs
+++ b/src/HealthChecks.AzureServiceBus/AzureServiceBusTopicHealthCheck.cs
@@ -36,17 +36,7 @@ namespace HealthChecks.AzureServiceBus
         {
             try
             {
-                var connectionKey = $"{ConnectionKey}_{_topicName}";
-                if (!ManagementClientConnections.TryGetValue(connectionKey, out var managementClient))
-                {
-                    managementClient = CreateManagementClient();
-                    if (!ManagementClientConnections.TryAdd(connectionKey, managementClient))
-                    {
-                        return new HealthCheckResult(context.Registration.FailureStatus,
-                            "New service bus administration client can't be added into dictionary.");
-                    }
-                }
-
+                var managementClient = ManagementClientConnections.GetOrAdd(ConnectionKey, _ => CreateManagementClient());
                 _ = await managementClient.GetTopicRuntimePropertiesAsync(_topicName, cancellationToken);
                 return HealthCheckResult.Healthy();
             }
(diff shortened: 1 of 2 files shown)


src/HealthChecks.AzureServiceBus/AzureServiceBusTopicHealthCheck.cs
```diff
@@ -36,17 +36,7 @@ public AzureServiceBusTopicHealthCheck(string endpoint, string topicName, TokenC
         {
             try
             {
-                var connectionKey = $"{ConnectionKey}_{_topicName}";
```
sungam3r, line 39: By the way, it seems a bug was here: `ConnectionKey` is property that already contains `_topicName`.

src/HealthChecks.AzureServiceBus/AzureServiceBusTopicHealthCheck.cs
```diff
@@ -36,17 +36,7 @@ public AzureServiceBusTopicHealthCheck(string endpoint, string topicName, TokenC
...
-                {
-                    managementClient = CreateManagementClient();
-                    if (!ManagementClientConnections.TryAdd(connectionKey, managementClient))
-                    {
-                        return new HealthCheckResult(context.Registration.FailureStatus,
-                            "New service bus administration client can't be added into dictionary.");
-                    }
-                }
-
+                var managementClient = ManagementClientConnections.GetOrAdd(ConnectionKey, _ => CreateManagementClient());
```
sungam3r, line 39: Please, also add discard parameter in `AzureServiceBusSubscriptionHealthCheck`.

src/HealthChecks.AzureServiceBus/AzureServiceBusTopicHealthCheck.cs
sungam3r: And consider to add `protected override string ConnectionKey => _connectionKey ??= ...` in both health checks to cache string in field though it is not necessary.

sungam3r (approved):

src/HealthChecks.AzureServiceBus/AzureServiceBusTopicHealthCheck.cs
davecluderay: Done.

sungam3r:
@CarlosLanderas @unaizorrilla  I see (not from settings because I have no access but from git history) that repo is configured for merge commits. Consider switching to squash to keep history clean:
изображение

Also for merge commits I have to fix commit message from "merge pull request from blablabla" to something more meaningful.

sungam3r:
I don't use any Azure healthecks, moreover I never used Azure as well :)

PR #913, opened 2022-02-20
Add editorconfig

sungam3r:
No code changes (except one bugfix for RavenDb), only formatting and docs; .editorconfig file was copied from [GraphQL.NET](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/.editorconfig) project.


src/HealthChecks.RavenDB/DependencyInjection/RavenDBHealthCheckBuilderExtensions.cs
```diff
@@ -39,7 +39,8 @@ public static class RavenDBHealthCheckBuilderExtensions
                 name ?? NAME,
                 sp => new RavenDBHealthCheck(options),
                 failureStatus,
-                tags));
+                tags,
+                timeout));
```
sungam3r, line 43: Bugfix here: `timeout` parameter was lost.

PR #915, opened 2022-02-21
Added support for using IServiceProvider when configuring Rabbitmq

isdaniel:
**What this PR does / why we need it**:

This pull request is supported for configuring the connection and option using an `IServiceProvider` similar to how the `SqlServerHealthCheck` supports this for a connection string & Uri. 

So instead of this:

```
var provider = services.BuildServiceProvider();
var settings = services.GetRequiredService<IOptions>();
```

We need to set string type for Rabbitmq connection from heartbeat before.

```
services.AddHealthChecks().AddRabbitMQ("Connectionstring");
```

You can now do something like the code below.

```
//for connection string
services.AddHealthChecks().AddRabbitMQ(sp => services.GetRequiredService<IOptions>().Value.RabbitmqConnectionString);

//for uri string
services.AddHealthChecks().AddRabbitMQ(sp => services.GetRequiredService<IOptions>().Value.RabbitmqConnectionString);
```

**Special notes for your reviewer**:

This is based on the connectionStringFactory implementation in the SqlServerHealthCheck.

I had added some unit tests & functional tests for this feature, I passed by this picture below

image

**Does this PR introduce a user-facing change?**:

Please make sure you've completed the relevant tasks for this PR, out of the following list:

It extends but does not break the existing options for configuring a RabbitMQ HealthCheck

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [x] End-to-end tests passing
- [ ] Extended the documentation
- [ ] Provided sample for the feature

PR #916, opened 2022-02-23
Add dotnet-format to CI

sungam3r:
I found it pretty tiring to fix formatting style and other stuff (example: #915).


src/HealthChecks.ArangoDb/README.md
```diff
@@ -14,17 +14,17 @@ With all of the following examples, you can additionally add the following param
 
 ### Basic
 
-```cs
```
sungam3r, line 17: `cs` is wrong moniker, code highlighting does not work

sungam3r:
This is an example of changes I would like to do with codebase, i.e. all other health checks. For a long time I have been using this utility in other OSS projects (and at my work as well) and it has proven itself well.

PR #925, opened 2022-02-23
Trigger CI on changes in directory.Build.props/targets

sungam3r:


sungam3r:
Keep all packages versions in one place was not the best idea with such CI organization.

PR #958, opened 2022-02-27
Add DiskStorageOptions.CheckAllDrives

### LINKED ISSUE #375: Usability concern with DiskStorageHealthCheck
When using `AddDiskStorageHealthCheck`, `setup` argument allow us to add several drive limit  checks through `AddDrive`:

``` csharp
healthCheckBuilder.AddDiskStorageHealthCheck(diskOptions =>
{
    diskOptions.AddDrive(driveName: "c:\\", 100);
    diskOptions.AddDrive(driveName: "f:\\", 300);
});
```

but `DiskStorageHealthCheck.CheckHealthAsync` implementation stops from the first non-compliance ([source code](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/55a2915506fc2c9c3c8eda639913d55b795b1959/src/HealthChecks.System/DiskStorageHealthCheck.cs#L32)).

In the case where multiple disks have exceeded the threshold, only first one is reported to `ResponseWriter` provided to `UseHealthChecks`.

It would be desirable to be able to report each non-conformance (for exemple using multiple lines in the description) rather than let an administrator discover them in several tries.

This behavior would also be aligned with the general behavior where each registered `IHealthCheck` are checked regardless of the result of the previous one.

This could be made opt-in with a bool CheckAllDrives in `DiskStorageOptions`

I know I can use:
``` csharp
healthCheckBuilder.AddDiskStorageHealthCheck(
    setup: diskOptions =>diskOptions.AddDrive(driveName: "c:\\", 100),
    name: "diskstorage " +  "c:\\");

healthCheckBuilder.AddDiskStorageHealthCheck(
    setup: diskOptions =>diskOptions.AddDrive(driveName: "f:\\", 00),
    name: "diskstorage " +  "f:\\");
```
but the hierarchical approach seems to be more elegant than flat one.

sungam3r:
Fixes #375. Also optimizes `DiskStorageHealthCheck` a bit.

### DIFF
--- a/src/HealthChecks.System/DiskStorageHealthCheck.cs
+++ b/src/HealthChecks.System/DiskStorageHealthCheck.cs
@@ -11,48 +11,49 @@ namespace HealthChecks.System
             _options = options ?? throw new ArgumentNullException(nameof(options));
         }
 
+        /// <inheritdoc/>
         public Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
         {
             try
             {
-                var configuredDrives = _options.ConfiguredDrives.Values;
+                var configuredDrives = _options.ConfiguredDrives;
+                List<string>? errors = null;
 
-                foreach (var (DriveName, MinimumFreeMegabytes) in configuredDrives)
+                if (configuredDrives.Count > 0)
                 {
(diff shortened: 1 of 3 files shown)


sungam3r2022:
Superseded by #987

PR #911, opened 2022-02-18
Update UIHealthReport.cs

mipa87:
Fix null Entries in _UIHealthReport_

**What this PR does / why we need it**:
Fixes the NullReferenceException when accessing null Entries in _UIHealthReport_. For example in _HealthReportExtensions_.

**Which issue(s) this PR fixes**:

~~Please reference the issue this PR will close: #_[issue number]_~~

**Special notes for your reviewer**:
Occurs when _HealthCheckReportCollector.GetHealthReport_ successfully deserializes the response, but in the requested URI is another API, which provide only "status" property and "entries" are omitted:
```json
{"title":"NotFound","status":404,"detail":"NotFound"}
```
 
**Does this PR introduce a user-facing change?**:
No

Please make sure you've completed the relevant tasks for this PR, out of the following list:

- [x] Code compiles correctly
- [ ] Created/updated tests
- [#] Unit tests passing
- [ ] End-to-end tests passing
- [ ] Extended the documentation
- [ ] Provided sample for the feature

Issue #862, opened 2021-11-23
Null ref exception on UIHealthReport.Entries


BartNetJS:
**What happened**: 
An api is protected with oauth and sitting behind azure APIM.
The response from APIM on an unauthorized call is a JSON response like this:
{
    "statusCode": 401,
    "message": "Access denied due to missing subscription key. Make sure to include subscription key when making requests to an API."
}

The extensions method HttpResponseMessageExtensions.As will deserialize is to a UIHealthReport with a null entries property.
And this causes a null ref exception in the extension method HealthReportExtensions.ToExecutionEntries

**What you expected to happen**:
An UIHealthReport with valid entries that gives more insights on the issue

**Source code sample**:

I suggest the following:
To avoid null ref exception:
![image](https://user-images.githubusercontent.com/4865355/143004388-f2bdbbce-5f37-44bf-b6c0-c086953fe8a3.png)

and to make the result on errors more verbose:
![image](https://user-images.githubusercontent.com/4865355/143004608-b67b8198-bcd8-42ba-a168-1cb02d66e829.png)

What do you think?

This need to be changed:
![image](https://user-images.githubusercontent.com/4865355/143005246-555fd377-0d08-42c9-b5cc-db5421b6c787.png)
But need to convert it back to TContent and I could not find a good way to do so.
Also the where TContent : UIHealthReport was needed to accomplish it in my debugging session, of course not how it should be.
Input appreciated :-)

carlosrecuero:
Hi @BartNetJS , could you please verify if version 6.0.3 of UI (https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/911), solves your issue?

sungam3r:
Closed as resolved due to lack of feedback.

PR #975, opened 2022-03-01
Use System.Text.Json to deserialize health report response

mipa87:
Use System.Text.Json to deserialize health report response in HealthCheckReportCollector.cs.

See [https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpContentJsonExtensions.cs](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpContentJsonExtensions.cs)

Better handling of bad responses
- when State of response contains bad value, e.g. 404
- when response body is empty

Please reference the issue this PR will close:
#_[862]_

Or it can be like this, but there is check for EnsureSuccessStatusCode() which fails for unhealthy status (returns 503), so I think this is not the right solution:
```csharp
return await _httpClient.GetFromJsonAsync<UIHealthReport>(absoluteUri, new JsonSerializerOptions(JsonSerializerDefaults.Web)
{
    Converters = {
        new JsonStringEnumConverter(null, false)
    }
});
```
See [https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.cs](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.cs)

No

- [x] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


sungam3r:
Is it really necessary to change all those `launchsettings` ?

src/HealthChecks.UI.Core/UIHealthReport.cs
```diff
@@ -9,22 +9,20 @@ namespace HealthChecks.UI.Core
      */
     public class UIHealthReport
```
sungam3r, line 10: 3 breaking changes here - you removed two setters and added new parameter into ctor. Is it done for proper deserialization? Will it work without these changes?

sungam3r:
> Please reference the issue this PR will close:
> #[862]

#862 was fixed in #911

sungam3r:
> Better handling of bad responses

    when State of response contains bad value, e.g. 404
    when response body is empty

What do you mean _better_ ?

src/HealthChecks.UI/Core/HostedService/HealthCheckReportCollector.cs
```diff
@@ -94,9 +93,13 @@ private async Task<UIHealthReport> GetHealthReportAsync(HealthCheckConfiguration
                 var absoluteUri = GetEndpointUri(configuration);
 
                 using var response = await _httpClient.GetAsync(absoluteUri, HttpCompletionOption.ResponseHeadersRead);
-                //response.EnsureSuccessStatusCode(); TODO: add or not ?
 
-                return await response.As<UIHealthReport>();
+                return await response.Content.ReadFromJsonAsync<UIHealthReport>(new JsonSerializerOptions(JsonSerializerDefaults.Web)
+                {
+                    Converters = {
+                        new JsonStringEnumConverter(null, false)
```
sungam3r, line 100:
```suggestion
                        new JsonStringEnumConverter(namingPolicy: null, allowIntegerValues: true)

```

src/HealthChecks.UI/Core/HostedService/HealthCheckReportCollector.cs
```diff
@@ -94,9 +93,13 @@ private async Task<UIHealthReport> GetHealthReportAsync(HealthCheckConfiguration
                 var absoluteUri = GetEndpointUri(configuration);
 
                 using var response = await _httpClient.GetAsync(absoluteUri, HttpCompletionOption.ResponseHeadersRead);
-                //response.EnsureSuccessStatusCode(); TODO: add or not ?
```
sungam3r, line 97: Are you sure? What are your considerations?

src/HealthChecks.UI.Core/UIHealthReport.cs
mipa87: Yes, it will works without this changes.

src/HealthChecks.UI/Core/HostedService/HealthCheckReportCollector.cs
mipa87: Unhealthy status returns HTTP status code 503, when we use EnsureSuccessStatusCode() it will throws exception and we wouldn't be able to process the response correctly.

mipa87:
> Is it really necessary to change all those `launchsettings` ?

It's not necessary, I updated it when I tested it.

mipa87:
> > Better handling of bad responses
> 
> ```
> when State of response contains bad value, e.g. 404
> when response body is empty
> ```
> 
> What do you mean _better_ ?

When Status in response is 404, using Newtonsoft Json this value is propagated to the UII, where it crashes in some places.

src/HealthChecks.UI/Core/HostedService/HealthCheckReportCollector.cs
sungam3r: You're right, I forgot about it.

sungam3r:
> When Status in response is 404, using Newtonsoft Json this value is propagated to the UII, where it crashes in some places.

So now the error will occur earlier?

src/HealthChecks.UI.Core/UIHealthReport.cs
sungam3r: Then please revert this file so your changes were backward compatible.

mipa87: It is reverted.

sungam3r:
@carlosrecuero Are you OK to change all those launchsettings files?

sungam3r:
> Integer in a string (like "0") will be fine and an integer as a number (like 0) will fail if allowIntegerValues is false.

🤔 I didn't know. OK. Though "0"/"1" seems a bit weird and maybe I would prefer to forbid such "enum values". 

I'm fine to merge as is if @carlosrecuero approve.

sungam3r (approved):

sungam3r2022:
@carlosrecuero Merge as you see fit. I have no write access with this new account.

PR #983, opened 2022-03-02
feat: Added support for using IServiceProvider when configuring Oracle

baldutech:
**What this PR does / why we need it:
This pull request adds support for configuring the key vault URI and option using an IServiceProvider similar to how the SqlServerHealthCheck supports this for a connection string. This way the ASP.net core options pattern can be used to retrieve the key vault URI and/or keys and secrets.

So instead of this:
var provider = services.BuildServiceProvider();
var settings = services.GetRequiredService<IOptions>();

services.AddHealthChecks().AddOracle( settings.Value.OracleConnectionString);

You can now do something like this:

services.AddHealthChecks().AddOracle( sp => services.GetRequiredService<IOptions>().Value.OracleConnectionString);

**Which issue(s) this PR fixes:
none

Special notes for your reviewer:
This is based on the connectionStringFactory implementation in the SqlServerHealthCheck.

Does this PR introduce a user-facing change?:
It extends but does not break the existing options for configuring a OracleHealthCheck

 Code compiles correctly
 Created/updated tests
 Unit tests passing
 End-to-end tests passing
 Extended the documentation
 Provided sample for the feature


sungam3r2022:
@carlosrecuero Hi from @sungam3r 😄 . My account was locked by GitHub. Reason? Just see latest news and you'll understand. I send you a letter from my email. So I think that next 1-2 days I will somehow repair overall process working on GitHub.

sungam3r:
@baldutech I'm fine to review again and merge after fixing suggestions.

sungam3r (approved):

sungam3r:
Thanks.

carlosrecuero:
Published on package 6.0.3

PR #996, opened 2022-03-09
Adding an optional Func to SqlServerHealthCheck to allow adding an Ac…

juanp3053:
…cessToken Provider for the SqlConnection object

Currently the SqlServerHealthCheck only allows authentication to SQL server via User/Password, which limits the ways of using the Health Check to SQL DBs that might user other types of authentication, such as an Access Token (Azure AD). I created an optional parameter so that an access token provider can be added, hence being able to connect to a SQL database with a connection string that does not have a User/Password in it but does make use of Access Tokens to Authenticate.

Yes, There is a new optional parameter. If this parameter is not supplied by the user the functionality is backwards compatible.

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.SqlServer/DependencyInjection/SqlServerHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.SqlServer/DependencyInjection/SqlServerHealthCheckBuilderExtensions.cs
@@ -26,6 +26,7 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <param name="tags">A list of tags that can be used to filter sets of health checks. Optional.</param>
         /// <param name="timeout">An optional <see cref="TimeSpan"/> representing the timeout of the check.</param>
         /// <param name="beforeOpenConnectionConfigurer">An optional action executed before the connection is Open on the healthcheck.</param>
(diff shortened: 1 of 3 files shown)


sungam3r:
You can set token with `beforeOpenConnectionConfigurer `: 
```
            services.AddHealthChecks()
                .AddSqlServer(connectionstring, beforeOpenConnectionConfigurer: c => c.AccessToken = "blablabla");
```

juanp3053:
@sungam3r Correct, but that is after the connection has been instantiated. It my testing, it would not authenticate properly using "beforeOpenConnectionConfigurer"

sungam3r:
изображение

You added property initialization, `beforeOpenConnectionConfigurer` works the same. Both of them just work with created instance of `SqlConnection`. I do not see any difference between two options.

juanp3053:
You are very right. After further investigation I have found that the issue was mainly because of the use of an async method in the action. Such as
```
 services.AddHealthChecks()
                .AddSqlServer(connectionstring, beforeOpenConnectionConfigurer: async (c) => c.AccessToken = await GetMyToken());
```
I have found a way of avoiding to use async method. You can go ahead and scrap this PR, my apologies for wasting your time.

sungam3r:
OK.

PR #997, opened 2022-03-09
Add nullable annotations to UI projects

sungam3r:
Also:
- moves JsonSerializerOptions into static field in HealthCheckReportCollector
- adds `xunit.runner.json` file with settings allowing to track long running tests


sungam3r:
Regarding my last comment - it's worth to deal with hanging tests in separate PR.

Issue #1003, opened 2022-03-10
Redis (Sentinel): The underlying connection is not renew if it's broken 


JorTurFer:
**What happened**: If you are running this health check in a Redis HA scenario using Sentinel (in Kubernetes), it could happen that the master node doesn't exist anymore because any cluster operation. Even the instance is recreated, the attached IP changes, so the connection is still broken. 

There is [an issue](https://github.com/StackExchange/StackExchange.Redis/issues/1584) in [StackExchange.Redis](https://github.com/StackExchange/StackExchange.Redis)[](https://github.com/StackExchange/StackExchange.Redis) explaining this problem.

This produces that after this situation, the healtcheck never comes back because it tries to hit an unexisting IP address.

**What you expected to happen**:
I'd expect that after new master node is elected, the health check work again

**How to reproduce it (as minimally and precisely as possible)**:

1. Deploy a Redis Sentinel solution (for example using redis-operator). 
2. Create an app using this check and run it.
3. Kill master node.
4. Hit the healthchek endpoint

**Anything else we need to know?**:
We have solved this problem removing the connection from the dictionary if the check fails, just for creating a fresh connection for the next time in that case. Upgrading the Redis package to latest version is another good improvement for Sentinel support.

I'm willing to open a PR with it if you think that this is useful

sungam3r:
> Upgrading the Redis package to latest version is another good improvement for Sentinel support.

OK but as you said here https://github.com/StackExchange/StackExchange.Redis/issues/1584#issuecomment-1063801016 that does not work.

JorTurFer:
> > Upgrading the Redis package to latest version is another good improvement for Sentinel support.
> 
> OK but as you said here [StackExchange/StackExchange.Redis#1584 (comment)](https://github.com/StackExchange/StackExchange.Redis/issues/1584#issuecomment-1063801016) that does not work.

Yes, that's true. It is their recommendation and that's why I suggested it, but our solution has been removing the connection from the dictionary if the check fails.
> We have solved this problem removing the connection from the dictionary if the check fails, just for creating a fresh connection for the next time in that case.

JorTurFer:
Let me open the PR because it's 3 minutes change, and we could discuss over the code

PR #1004, opened 2022-03-10
Redis: Renew the underlying connection if there is an exception

JorTurFer:
**What this PR does / why we need it**: This PR solves the reconnection problem in Redis with Sentinel arch after master node is missing

**Which issue(s) this PR fixes**: Fixes https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/1003

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [x]  Provided sample for the feature

### DIFF
--- a/src/HealthChecks.Redis/RedisHealthCheck.cs
+++ b/src/HealthChecks.Redis/RedisHealthCheck.cs
@@ -63,6 +63,7 @@ namespace HealthChecks.Redis
             }
             catch (Exception ex)
             {
+                _connections.TryRemove(_redisConnectionString, out _);
                 return new HealthCheckResult(context.Registration.FailureStatus, exception: ex);
             }
         }
--- a/src/HealthChecks.Redis/RedisHealthCheck.cs
+++ b/src/HealthChecks.Redis/RedisHealthCheck.cs
@@ -63,7 +63,8 @@ namespace HealthChecks.Redis
             }
             catch (Exception ex)
             {
-                _connections.TryRemove(_redisConnectionString, out _);
+                _connections.TryRemove(_redisConnectionString, out var connection);
+                connection?.Dispose();
                 return new HealthCheckResult(context.Registration.FailureStatus, exception: ex);
             }
         }
--- a/src/HealthChecks.Redis/HealthChecks.Redis.csproj
+++ b/src/HealthChecks.Redis/HealthChecks.Redis.csproj
@@ -9,7 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="StackExchange.Redis" Version="2.2.88" />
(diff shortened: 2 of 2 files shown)


JorTurFer:
@sungam3r PTAL

src/HealthChecks.Redis/RedisHealthCheck.cs
```diff
@@ -63,6 +63,7 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
             }
             catch (Exception ex)
             {
+                _connections.TryRemove(_redisConnectionString, out _);
```
sungam3r, line 66: I doubt sbout `TryGetValue/TryAdd` pair above. See #903, #905 . Also note that `ConnectionMultiplexer` dictionary value implements `IDisposable` and should be disposed.

JorTurFer: I don't get the first part, sorry.
Would you like that, instead of using `TryGetValue/TryAdd/TryRemove` I update it to match the approach from the PR that you have posted?
Related with IDisposable you are right, I'll update my PR with that 😄

sungam3r: > Would you like that, instead of using TryGetValue/TryAdd/TryRemove I update it to match the approach from the PR that you have posted?

Yes. In fact that should be done in many places to get rid of race conditions.

JorTurFer: Thanks for the clarification ❤️ 
I'll update my PR with both changes

JorTurFer: I'm doing the change and I have faced with a problem using GetOrAdd, it hasn't got an async method. That pushes me to use something like this:
```csharp
var connection = _connections.GetOrAdd(_redisConnectionString, async key =>
{
    return ConnectionMultiplexer.Connect(key);
});                
```

My concern here is about doing sync calls inside the async context of the health check execution. Maybe I create a bigger problem with that than the original race condition with TryAdd/TryGet/TryRemove.
Any thoughts/suggestion @sungam3r ?

sungam3r: OK. Let's keep as is for now.

sungam3r (approved):

carlosrecuero:
Hi Jorge, I think I will release this package this weekend.

JorTurFer:
wow, so fast! Thanks a ton ❤️

carlosrecuero:
Published on package 6.0.3

JorTurFer:
❤️ ❤️ ❤️ ❤️

PR #1016, opened 2022-03-12
Dotnet CI rework

sungam3r:
- Adds nullable annotations to test projects so all projects are covered by nullable annotations.
- Adds `dotnet format` to test projects.
- Fixes some build warnings.
- Adds details on failed functional tests for Oracle and Solr.
- Adds TreatWarninsAsErrors=true option


sungam3r:
@carlosrecuero I need your help with annotations. I fixed majority of cases across codebase but I have trouble with the rest. I do not know the code well and in some cases was not confident in the correctness of changes. You may check out this branch, ~27 errors left. I suspect some warnings indicate that the code must be rewritten a bit. For example, I have seen a lot of `FirstOrDefault` method usages, when its result has not been checked on `null` anywhere.

sungam3r:
I set missing annotations, although in some cases it seems to me a more thoughtful rework of the code is required.

Issue #1048, opened 2022-03-27
Question about NuGet/setup-nuget


sungam3r:
I see that each CD workflow uses 
```yaml
  - name: setup nuget
      uses: NuGet/setup-nuget@v1.0.5
      env:
        ACTIONS_ALLOW_UNSECURE_COMMANDS: 'true'
      with:
        nuget-version: latest
```

After that it executes 
```yaml
 - name: Publish <NAME_HERE> Health Check nuget
      run: dotnet nuget push ./artifacts/AspNetCore.HealthChecks.NAME.*.nupkg -k ${{secrets.NUGET_API_KEY}} -s https://api.nuget.org/v3/index.json --skip-duplicate
```

but `dotnet nuget push` is a built-in dotnet CLI command. So why to bother with _setup nuget_ step in all CD workflows?
ping @carlosrecuero @unaizorrilla

sungam3r:
@carlosrecuero  Also I suggest to get rid of `--include-source --include-symbols` in all `pack` command. With embedded PDBs and SourceLink all that jumping over "debugger options" is unnecessary.

PR #1061, opened 2022-04-01
Run tests only for CI workflows

### LINKED ISSUE #1022: Why to tun tests for CD?
What is the reason to run tests for CD workflows? Assuming that CI occurs only on the `master` branch, this means that all the tests have already been completed earlier during CI workflows. I understand that in the general case you can run tests whenever you want but for CD purposes it may be a problem if some test fails for unknown/transient reason. I propose delete `dotnet test` commands from all CD workflows. This implies also that services docker settings may be removed as well. Only CI workflows need to bother with these settings to run tests.  @carlosrecuero @unaizorrilla What do you think?

sungam3r:
fixes #1022

Also adds `--no-restore` to `dotnet pack`.

### DIFF
--- a/.github/workflows/healthchecks_arangodb_cd.yml
+++ b/.github/workflows/healthchecks_arangodb_cd.yml
@@ -9,13 +9,6 @@ jobs:
     env:
       BUILD_CONFIG: Release
     runs-on: ubuntu-latest
-    services:
-      arangodb:
-        image: arangodb/arangodb:latest
-        ports:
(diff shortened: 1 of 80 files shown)


sungam3r:
@carlosrecuero If OK I would like to proceed with PR fixing #1048 .

carlosrecuero (approved):

carlosrecuero:
Yes, sure.

Issue #1171, opened 2022-05-23
HealthCheck.Publisher.Seq produces wrong API URI if endpoint has trailing slash


dccs-pwurzinger:
Please, fill the following sections to help us fix the issue

**What happened**:
Seq-publisher used a malformed endpoint URI to publish health check reports:

`HTTP POST http://localhost:5341//api/events/raw?apiKey=...`
(Please note the second slash after the port.)
This call resulted in a series of HTTP 404s being returned, obviously.

**What you expected to happen**:
`HTTP POST http://localhost:5341/api/events/raw?apiKey=...`
(Please note only one slash after the port.)

**How to reproduce it (as minimally and precisely as possible)**:
Configuring the Seq-publisher with an endpoint, that has a trailing slash, e.g.
```cs
 builder.Services.AddHealthChecks()
                 .AddSeqPublisher(options => options.Endpoint = "http://localhost:5341/");
```

**Anything else we need to know?**:
The reason that this happens is because the API endpoint is being generated by concattenating the Seq host with the API-Endpoint route in here:
https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/c60573659a5e415fbbf0fda8363ca798b32890f4/src/HealthChecks.Publisher.Seq/SeqPublisher.cs#L66

It should be an easy fix using `System.Uri` and/or `PathString` for that, that already checks for trailing slashes and so on. If you want, I can provide a PR fixing that?

**Environment**:

- .NET Core version: .NET Core 6 & ASP.NET Core 6
- Healthchecks version 6.0.2

**Workaround**
If somebody using version AspNetCore.Diagnostics.HealthChecks 6.0.2, this is the workaround we are using:
```cs
  private static string NormalizeBaseAddress(Uri seqServiceUri)
  {
      var seqServiceUriString = seqServiceUri.ToString();

      var hasTrailingBackslash = seqServiceUriString.EndsWith("/", StringComparison.OrdinalIgnoreCase);
      if (hasTrailingBackslash)
          return seqServiceUriString[..^1];
      return seqServiceUriString;
  }
```

sungam3r:
PR is welcome.

sungam3r:
fixed by #1304

dccs-pwurzinger:
Awesome, thanks! I didn't have the time to provide a PR, my apologies.

PR #1304, opened 2022-06-23
Fix seq trailing slash, respect cancellation token, make SeqOptions.ApiKey nullable

MrSmoke:
- Fixes a bug where the Seq publisher does not handle trailing slashes
- Fixed Seq publisher not respecting the cancellation token
- Throw exception on null/empty endpoint in `SeqOptions`
- `SeqOptions.ApiKey` is now nullable to handle cases where no api key is required
- Fixed `seq_publisher_registration_should.add_healthcheck_when_properly_configured` options configured incorrectly

#1171

No

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.Publisher.Seq/SeqOptions.cs
+++ b/src/HealthChecks.Publisher.Seq/SeqOptions.cs
@@ -6,7 +6,7 @@ namespace Microsoft.Extensions.DependencyInjection
     {
         public string Endpoint { get; set; } = null!;
 
-        public string ApiKey { get; set; } = null!;
+        public string? ApiKey { get; set; }
 
         public SeqInputLevel DefaultInputLevel { get; set; }
     }
--- a/src/HealthChecks.Publisher.Seq/SeqPublisher.cs
+++ b/src/HealthChecks.Publisher.Seq/SeqPublisher.cs
@@ -9,7 +9,6 @@ namespace HealthChecks.Publisher.Seq
 {
     public class SeqPublisher : IHealthCheckPublisher
     {
-
         private readonly SeqOptions _options;
         private readonly Func<HttpClient> _httpClientFactory;
 
@@ -33,6 +32,8 @@ namespace HealthChecks.Publisher.Seq
                     break;
             }
 
+            string? assemblyName = Assembly.GetEntryAssembly()?.GetName().Name;
+
             var events = new RawEvents
             {
                 Events = new RawEvent[]
@@ -40,12 +41,12 @@ namespace HealthChecks.Publisher.Seq
                     new RawEvent
                     {
                         Timestamp = DateTimeOffset.UtcNow,
-                        MessageTemplate = $"[{Assembly.GetEntryAssembly()?.GetName().Name} - HealthCheck Result]",
+                        MessageTemplate = $"[{assemblyName} - HealthCheck Result]",
                         Level = level.ToString(),
                         Properties = new Dictionary<string, object?>
                         {
                             { nameof(Environment.MachineName), Environment.MachineName },
-                            { nameof(Assembly), Assembly.GetEntryAssembly()?.GetName().Name },
+                            { nameof(Assembly), assemblyName },
                             { "Status", report.Status.ToString() },
                             { "TimeElapsed", report.TotalDuration.TotalMilliseconds },
                             { "RawReport" , JsonConvert.SerializeObject(report)}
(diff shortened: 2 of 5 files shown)


src/HealthChecks.Publisher.Seq/SeqPublisher.cs
```diff
@@ -54,29 +57,40 @@ public async Task PublishAsync(HealthReport report, CancellationToken cancellati
                 }
             };
 
-            await PushMetricsAsync(JsonConvert.SerializeObject(events));
+            await PushMetricsAsync(JsonConvert.SerializeObject(events), cancellationToken);
         }
 
-        private async Task PushMetricsAsync(string json)
+        private async Task PushMetricsAsync(string json, CancellationToken token)
```
sungam3r, line 63:
```suggestion
        private async Task PushMetricsAsync(string json, CancellationToken cancellationToken)
```

test/HealthChecks.Publisher.Seq.Tests/Functional/SeqPublisherTests.cs
```diff
@@ -0,0 +1,48 @@
...
+        using var handler = new MockClientHandler();
+        HttpClient HttpClientFactory() => new(handler);
+
+        var testReport = new HealthReport(new Dictionary<string, HealthReportEntry>(), TimeSpan.Zero);
+
+        // Create publisher and publish
+        var publisher = new SeqPublisher(HttpClientFactory, options);
+        await publisher.PublishAsync(testReport, CancellationToken.None);
+
+        Assert.NotNull(handler.Request);
```
sungam3r, line 28: Please use FluentAssertions extension methods instead of Assert.

MrSmoke: done

src/HealthChecks.Publisher.Seq/SeqPublisher.cs
MrSmoke: done

src/HealthChecks.Publisher.Seq/SeqOptions.cs
```diff
@@ -6,7 +6,7 @@ public class SeqOptions
     {
         public string Endpoint { get; set; } = null!;
 
-        public string ApiKey { get; set; } = null!;
+        public string? ApiKey { get; set; }
```
MrSmoke, line 9: I had a look at the official API client for seq and it seems that the `ApiKey` can be null so it probably makes sense for it to be null here too?

https://github.com/datalust/seq-api/blob/dev/src/Seq.Api/Client/SeqApiClient.cs#L75=

sungam3r: yes

src/HealthChecks.Publisher.Seq/SeqPublisher.cs
```diff
@@ -54,29 +57,50 @@ public async Task PublishAsync(HealthReport report, CancellationToken cancellati
...
+            if (string.IsNullOrEmpty(options.Endpoint))
+                throw new ArgumentNullException(nameof(options.Endpoint));
+
+            var uriBuilder = new UriBuilder(options.Endpoint)
+            {
+                Path = "/api/events/raw",
+            };
+
+            // Add api key if supplied
+            if (!string.IsNullOrWhiteSpace(options.ApiKey))
```
sungam3r, line 98:
```suggestion
            if (!string.IsNullOrEmpty(options.ApiKey))
```

or just `if (options.ApiKey != null)`

test/HealthChecks.Publisher.Seq.Tests/DependencyInjection/RegistrationTests.cs
```diff
@@ -9,12 +9,11 @@ public void add_healthcheck_when_properly_configured()
             services
                 .AddHealthChecks()
                 .AddSeqPublisher(setup =>
-                    new SeqOptions()
```
sungam3r, line 12: 👍 funny test fix

src/HealthChecks.Publisher.Seq/SeqPublisher.cs
sungam3r: Ready to merge after this fix.

sungam3r (approved):

PR #1345, opened 2022-07-14
Add multiple NpgSql db support

vadrsa:

- [ x] Code compiles correctly
- [ x] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/src/HealthChecks.NpgSql/DependencyInjection/NpgSqlHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.NpgSql/DependencyInjection/NpgSqlHealthCheckBuilderExtensions.cs
@@ -68,12 +68,12 @@ namespace Microsoft.Extensions.DependencyInjection
             {
                 throw new ArgumentNullException(nameof(connectionStringFactory));
             }
-
-            builder.Services.AddSingleton(sp => new NpgSqlHealthCheck(connectionStringFactory(sp), healthQuery, connectionAction));
+            var finalName = name ?? NAME;
+            builder.Services.AddTransient(sp => new NpgSqlHealthCheck(finalName, connectionStringFactory(sp), healthQuery, connectionAction));
 
             return builder.Add(new HealthCheckRegistration(
-                name ?? NAME,
-                sp => sp.GetRequiredService<NpgSqlHealthCheck>(),
+                finalName,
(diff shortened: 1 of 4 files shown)


src/HealthChecks.NpgSql/DependencyInjection/NpgSqlHealthCheckBuilderExtensions.cs
```diff
@@ -68,12 +68,12 @@ public static class NpgSqlHealthCheckBuilderExtensions
...
-
-            builder.Services.AddSingleton(sp => new NpgSqlHealthCheck(connectionStringFactory(sp), healthQuery, connectionAction));
+            var finalName = name ?? NAME;
+            builder.Services.AddTransient(sp => new NpgSqlHealthCheck(finalName, connectionStringFactory(sp), healthQuery, connectionAction));
 
             return builder.Add(new HealthCheckRegistration(
-                name ?? NAME,
-                sp => sp.GetRequiredService<NpgSqlHealthCheck>(),
+                finalName,
+                sp => sp.GetServices<NpgSqlHealthCheck>().Single(x => x.Name == finalName),
```
sungam3r, line 76: It will blow up existing code like that one:
```c#
builder.AddNpgSql(constr);
builder.AddNpgSql(constr);
```

vadrsa: Yeah, we can keep the old behavior for multiple unnamed checks by doing `.Last()` here.

sungam3r:
Honestly, I doubt the correctness of this path in terms of design. It seems to me that in a situation where the application uses several db connections, only it itself can decide which connection is good and which is bad in every situation. So the custom healthcheck seems more suitable here.

sungam3r:
The difference is not in db provider (sql server, mongodb or postgres) but in design of checking itself. HealtChecks are designed to be additive IMO. Here you propose to add a check that effectively selects other check to execute. That behaviour is not specific to postgress and can be (and should be IMO) considered as a general-purpose design. Ping @carlosrecuero  for comments.

sungam3r:
@vadrsa Thanks for contribution. As I said earlier I doubt the correctness of this path in terms of design so I close this PR.

PR #1358, opened 2022-07-25
Fix AzureKeyVaultHealthChecksBuilderExtensions

### LINKED ISSUE #1356: AzureKeyVault InvalidOperationException: Operations that change non-concurrent collections must have exclusive access
**What happened**:
I have health checks configured in Azure to run automatically as part of Azure's built in integration with app service. I have intermittent issue where the health checks are failing with a 500 error (this seems to start when a app service instance is restarted, and doesn't stop until the app service is restarted again). I have narrowed this down to the AzureKeyVault health check.

image

This is pointing me to where I configure the key vault health check and right at the HashSet collection that is used to add the secrets.

For this error not to occur

I have not found a way to reproduce the issue at this time.

**Source code sample**:
This is what my code looks like. I have a collection of secrets that I store in a list that I add via foreach loop. 
`
// Collection of 9 secrets that I add
public static readonly List<string> Secrets = new() { "1", "2", "3", "4", "5", "6", "7", "8", "9" }; 

services.AddHealthChecks()
					.AddAzureKeyVault(new Uri(uri), new DefaultAzureCredential(),
						options => {
							// secrets
							foreach (string secret in secrets) {
								options.AddSecret(secret));
							}
						}, "KeyVault_Check");`

I'm considering just adding only a single secret instead of multiple (knowing more about health checks in general it doesn't seem worth testing beyond a single secret's existence). 

**Environment**:

- .NET Core version 6
- Healthchecks version 6.0.2 of AspNetCore.HealthChecks.AzureKeyVault
- Operative system: Azure windows app service
- Others:

sungam3r:
Fixes #1356

### DIFF
--- a/src/HealthChecks.AzureKeyVault/AzureKeyVaultOptions.cs
+++ b/src/HealthChecks.AzureKeyVault/AzureKeyVaultOptions.cs
@@ -5,15 +5,15 @@ namespace HealthChecks.AzureKeyVault
     /// </summary>
     public class AzureKeyVaultOptions
     {
-        internal HashSet<string> _secrets = new HashSet<string>();
+        internal HashSet<string> _secrets = new();
 
         internal IEnumerable<string> Secrets => _secrets;
 
-        internal HashSet<string> _keys = new HashSet<string>();
+        internal HashSet<string> _keys = new();
 
         internal IEnumerable<string> Keys => _keys;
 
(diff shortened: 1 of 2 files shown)


ddoorn (approved):
Thank you for the PR to resolve the issue. I'm looking forward to giving this a try once it is completed and available in nuget.

sungam3r:
https://www.nuget.org/packages/AspNetCore.HealthChecks.AzureKeyVault/6.0.3

Issue #1371, opened 2022-08-03
Leverage Existing Azure Service Clients Using Dependency Injection


wsugarman:
**What would you like to be added**: 
Overloads to the various Azure service health check registration methods that re-use an existing service client from the service container.

**Why is this needed**:
Many users of the Azure client SDKs register their clients as a service in their service containers to easily centralize their creation; perhaps the clients are created manually or are loaded from a configuration. For example, users may include something similar to the following in their ASP.NET Core applications:

```csharp
public IConfiguration Configuration { get; }

// ...

services.AddAzureClients(builder => builder.AddBlobServiceClient(Configuration));
```

However, the health checks (at least the ones I've looked at) do not have a registration overload for fetching the client from an `IServiceProvider`. Therefore, users must re-specify all of the settings used to configure their clients.

Nisden:
The clean up that happen in the pull request #1405 broke the changes made to fix #783 as using `GetPropertiesAsync` for checking blob storage access is not possible when using an Azure AD account that only have "Blob Storage Contributor" as permission instead of "Blob Storage Owner"

wsugarman:
> The clean up that happen in the pull request #1405 broke the changes made to fix #783 as using `GetPropertiesAsync` for checking blob storage access is not possible when using an Azure AD account that only have "Blob Storage Contributor" as permission instead of "Blob Storage Owner"

Oh! My mistake. I'll put out another PR to fix this and add a comment so that hopefully the next person doesn't undo it like me.

PR #1372, opened 2022-08-03
Optionally Use Azure Storage Clients from IServiceProvider

wsugarman:
Adds new health check registration overloads that reuse an existing client from the `IServiceProvider` instead of creating new ones. Includes the following changes:
- Add new registration methods for reusing configured clients
- Add new ctors to each health check that accepts the client
- Standardize the health check logic across the Azure Blob, File Share, and Queue services
- Centralize client caching

**Which issue(s) this PR fixes**: Relates to https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/1371

**Does this PR introduce a user-facing change?**: Yes. New APIs

- [X] Code compiles correctly
- [X] Created/updated tests
- [X] Unit tests passing
- [X] Provided sample for the feature

Sample
```csharp
services
    .AddHealthChecks()
    .AddAzureBlobStorage(o => o.ContainerName = "sample-container", name: "sample");
```


test/HealthChecks.AzureStorage.Tests/DependencyInjection/AzureBlobStorageRegistrationTests.cs
```diff
@@ -7,7 +7,7 @@ public void add_health_check_when_properly_configured()
         {
             var services = new ServiceCollection();
             services.AddHealthChecks()
-                .AddAzureBlobStorage("the-connection-string");
+                .AddAzureBlobStorage(connectionString: "the-connection-string");
```
wsugarman, line 10: As you can see, these new methods cause ambiguity which is not desirable. I'm thinking about how to introduce this more cleanly, especially for existing users.

build/dependencies.props
```diff
@@ -14,7 +14,7 @@
     <HealthCheckAWSSystemsManager>6.0.0</HealthCheckAWSSystemsManager>
     <HealthCheckAzureDigitalTwin>6.0.0</HealthCheckAzureDigitalTwin>
     <HealthCheckAzureServiceBus>6.0.4</HealthCheckAzureServiceBus>
-    <HealthCheckAzureStorage>6.0.4</HealthCheckAzureStorage>
+    <HealthCheckAzureStorage>6.0.5</HealthCheckAzureStorage>
```
wsugarman, line 17: New APIs may warrant a minor version bump. Although, perhaps we could re-use an existing API

src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
```diff
@@ -7,35 +7,34 @@ namespace HealthChecks.AzureStorage
 {
     public class AzureBlobStorageHealthCheck : IHealthCheck
     {
-        private readonly TokenCredential? _azureCredential;
-        private readonly Uri? _blobServiceUri;
-
-        private readonly string? _connectionString;
+        private readonly Func<BlobServiceClient> _clientFactory;
```
wsugarman, line 10: I want to preserve the existing functionality, so I tried encapsulating it in a factory member. Alternatively, we could probably create two derived classes that fetch their respective clients differently.

build/dependencies.props
sungam3r: I'm not sure about versioning model here (patch or minor). ping @carlosrecuero

sungam3r: Personally I'm fine either way.

src/HealthChecks.AzureStorage/AzureFileShareHealthCheck.cs
```diff
@@ -6,26 +6,39 @@ namespace HealthChecks.AzureStorage
...
-            _shareName = shareName;
+            if (connectionString == null)
+            {
+                throw new ArgumentNullException(nameof(connectionString));
+            }
+
+            _clientFactory = () => _shareClientsHolder.GetOrAdd($"{connectionString}{shareName}", _ => new ShareClient(connectionString, shareName));
+        }
+
+        public AzureFileShareHealthCheck(ShareClient shareClient)
```
sungam3r, line 23: I suggest to add one more `AzureFileShareHealthCheck` ctor with `Func<ShareClient>` argument and overloads to  `AzureStorageHealthCheckBuilderExtensions` with `ShareClient` and `IServiceProvider`.

src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
sungam3r: Factory method is fine.

test/HealthChecks.AzureStorage.Tests/DependencyInjection/AzureBlobStorageRegistrationTests.cs
sungam3r: Compile-time error in any case is more preferable then run-time error.

wsugarman: The new method overloads now accept a delegate for configuring the options, so these method signatures are no longer ambiguous!

build/dependencies.props
wsugarman: I've bumped it to minor for now, but I noticed that no other package has done so haha

src/HealthChecks.AzureStorage/AzureFileShareHealthCheck.cs
wsugarman: I've changed this such that each health check simply has a member representing the service client. How do you feel about this instead of the factories? Was there a reason I missed for resolving the clients within the method as opposed to the ctor? To accommodate the change, I did have to update the unit tests with real connection strings that could be parsed.

If that's ok, do we need the static client caching?

src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
```diff
@@ -1,85 +1,46 @@
-using System.Collections.Concurrent;
 using Azure.Core;
 using Azure.Storage.Blobs;
 using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Microsoft.Extensions.Options;
 
 namespace HealthChecks.AzureStorage
 {
     public class AzureBlobStorageHealthCheck : IHealthCheck
```
wsugarman, line 8: If I could propose a breaking change for the eventual 7.0.0 version, it would be making these health check classes internal, just so we don't have to expose their classes and ctors

sungam3r: This contradicts the general HC design from this repo.

src/HealthChecks.AzureStorage/AzureQueueStorageHealthCheck.cs
```diff
@@ -1,79 +1,46 @@
...
     {
-        private readonly string? _connectionString;
-        private readonly string? _queueName;
-
-        private readonly TokenCredential? _azureCredential;
-        private readonly Uri? _queueServiceUri;
-
-        private static readonly ConcurrentDictionary<string, QueueServiceClient> _queueClientsHolder = new();
+        private readonly QueueServiceClient _queueServiceClient;
+        private readonly IOptionsSnapshot<QueueStorageHealthCheckOptions> _options;
```
sungam3r, line 11: `IOptionsSnapshot` usages and overall design look overcomplicated. No one HC from this repo uses `IOptionsSnapshot` and even `IOptions`. I have a strong feeling that it is done so intentionally.

sungam3r: Look at other HCs and extension methods that configure them. The preferred way is to use `Action` or `Func` delegates with `IServiceProvider`.

wsugarman: I've removed IOptions and used the aforementioned pattern instead

src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
wsugarman: Then certainly we can leave it `public`

src/HealthChecks.AzureStorage/BlobStorageHealthCheckOptions.cs
```diff
@@ -0,0 +1,16 @@
+namespace HealthChecks.AzureStorage;
+
+/// <summary>
+/// Represents a collection of settings that configure Azure Storage Blob Service health checks.
+/// </summary>
+public sealed class BlobStorageHealthCheckOptions
```
sungam3r, line 6:
```suggestion
public sealed class AzureBlobStorageHealthCheckOptions
```

src/HealthChecks.AzureStorage/BlobStorageHealthCheckOptions.cs
```diff
@@ -0,0 +1,16 @@
+namespace HealthChecks.AzureStorage;
+
+/// <summary>
+/// Represents a collection of settings that configure Azure Storage Blob Service health checks.
```
sungam3r, line 4:
```suggestion
/// Represents a collection of settings that configure <see cref="AzureBlobStorageHealthCheck">Azure Storage Blob Service health check</see>.
```

The same suggestion for other options.

test/HealthChecks.AzureStorage.Tests/DependencyInjection/AzureFileShareStorageRegistrationTests.cs
```diff
@@ -7,7 +7,7 @@ public void add_health_check_when_properly_configured()
         {
             var services = new ServiceCollection();
             services.AddHealthChecks()
-                .AddAzureFileShare("the-connection-string");
+                .AddAzureFileShare("FileEndpoint=https://unit-test.file.core.windows.net");
```
sungam3r, line 10: Is such change nessessary for test purposes? I don't mind to change fake connection string to some real value but I am interested in whether this change is caused by an earlier creation of the client and connection string parsing.

test/HealthChecks.AzureStorage.Tests/HealthChecks.AzureStorage.approved.txt
```diff
@@ -2,29 +2,50 @@ namespace HealthChecks.AzureStorage
 {
     public class AzureBlobStorageHealthCheck : Microsoft.Extensions.Diagnostics.HealthChecks.IHealthCheck
     {
+        public AzureBlobStorageHealthCheck(Azure.Storage.Blobs.BlobServiceClient blobServiceClient, Microsoft.Extensions.Options.IOptionsSnapshot<HealthChecks.AzureStorage.BlobStorageHealthCheckOptions> options) { }
```
sungam3r, line 5: Forgot to update approval file.

wsugarman: Whoops! Good catch!

test/HealthChecks.AzureStorage.Tests/DependencyInjection/AzureFileShareStorageRegistrationTests.cs
wsugarman: You're right -- it's because we create the client in the ctor now instead of within the `CheckHealthAsync` method. The underlying service client parses the connection string and will throw if it's not formatted properly. We still don't need the corresponding service to exist, but the value must parse correctly.

sungam3r: ok

src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheckOptions.cs
```diff
@@ -0,0 +1,17 @@
+namespace HealthChecks.AzureStorage;
+
+/// <summary>
+/// Represents a collection of settings that configure
+/// <see cref="AzureBlobStorageHealthCheck">Azure Storage Blob Service health checks</see>.
```
sungam3r, line 5:
```suggestion
/// <see cref="AzureBlobStorageHealthCheck">Azure Storage Blob Service health check</see>.
```

src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
```diff
@@ -7,79 +6,38 @@ namespace HealthChecks.AzureStorage
...
-
-                    await containerClient.GetPropertiesAsync(cancellationToken: cancellationToken);
-                }
-
-                return HealthCheckResult.Healthy();
-            }
-            catch (Exception ex)
-            {
-                return new HealthCheckResult(context.Registration.FailureStatus, exception: ex);
-            }
```
sungam3r, line 61: Why did you remove this? This is a "standard" pattern for all HCs across repo.

sungam3r: The same note for other HCs.

src/HealthChecks.AzureStorage/AzureFileShareHealthCheck.cs
```diff
@@ -1,39 +1,36 @@
...
         }
 
         public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
         {
-            try
-            {
-                var shareClient = _shareClientsHolder.GetOrAdd($"{_connectionString}{_shareName}", _ => new ShareClient(_connectionString, _shareName));
+            await _shareServiceClient.GetPropertiesAsync(cancellationToken);
 
-                if (!await shareClient.ExistsAsync(cancellationToken))
```
sungam3r, line 26: Why did you remove `ExistsAsync`?

src/HealthChecks.AzureStorage/AzureQueueStorageHealthCheck.cs
```diff
@@ -7,73 +6,38 @@ namespace HealthChecks.AzureStorage
...
-            try
-            {
-                var queueServiceClient = GetQueueServiceClient();
-                var serviceProperties = await queueServiceClient.GetPropertiesAsync(cancellationToken);
-
-                if (!string.IsNullOrEmpty(_queueName))
-                {
-                    var queueClient = queueServiceClient.GetQueueClient(_queueName);
-
-                    if (!await queueClient.ExistsAsync(cancellationToken))
```
sungam3r, line 42: Again deleted.

src/HealthChecks.AzureStorage/AzureQueueStorageHealthCheckOptions.cs
```diff
@@ -0,0 +1,17 @@
+namespace HealthChecks.AzureStorage;
+
+/// <summary>
+/// Represents a collection of settings that configure
+/// <see cref="AzureQueueStorageHealthCheck">Azure Storage Queue Service health checks</see>.
```
sungam3r, line 5:
```suggestion
/// <see cref="AzureQueueStorageHealthCheck">Azure Storage Queue Service health check</see>.
```

src/HealthChecks.AzureStorage/DependencyInjection/AzureStorageHealthCheckBuilderExtensions.cs
```diff
@@ -76,22 +78,59 @@ public static class AzureStorageHealthCheckBuilderExtensions
...
+        /// <param name="configureOptions">Delegate for configuring the health check. Optional.</param>
+        /// <param name="name">The health check name. Optional. If <see langword="null"/> the type name 'azureblob' will be used for the name.</param>
+        /// <param name="failureStatus">
+        /// The <see cref="HealthStatus"/> that should be reported when the health check fails. Optional. If <see langword="null"/> then
+        /// the default status of <see cref="HealthStatus.Unhealthy"/> will be reported.
+        /// </param>
+        /// <param name="tags">A list of tags that can be used to filter sets of health checks. Optional.</param>
+        /// <param name="timeout">An optional <see cref="TimeSpan"/> representing the timeout of the check.</param>
+        /// <returns>The specified <paramref name="builder"/>.</returns>
+        public static IHealthChecksBuilder AddAzureBlobStorage(
```
sungam3r, line 104: Please also add overload without service provider:
```
  public static IHealthChecksBuilder AddAzureBlobStorage(
            Action<AzureBlobStorageHealthCheckOptions>? configureOptions = default,
            string? name = default,
            TimeSpan? timeout = default)
        => builder.AddAzureBlobStorage((_, options) => configureOptions?.Invoke(options), name, failureStatus, tags, timeout);
```

The same note for other HCs.

src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
wsugarman: I mentioned it briefly in the PR description, but this try/catch block is superfluous. The health check is invoked from the underlying `DefaultHealthCheckService` and that class will catch any unhandled exceptions and return unhealthy. The only difference is that it will use the exception's message as the message for the result. If we don't want that behavior, I can change it back. I just thought for simplicity's sake we could remove it

src/HealthChecks.AzureStorage/AzureFileShareHealthCheck.cs
wsugarman: If the container doesn't exist, `GetProperties` will throw a `RequestFailedException` with a 404 status and subsequently that 404 will be converted to a failed health check. That way, for the typical "healthy" code path, there are only 2 remote calls made to the Azure Storage service instead of 3.

test/HealthChecks.AzureStorage.Tests/DependencyInjection/AzureQueueStorageRegistrationTests.cs
```diff
@@ -1,92 +1,94 @@
...
+        public void add_health_check_with_client_from_service_provider(string? queueName, string? registrationName, HealthStatus? failureStatus)
         {
-            var services = new ServiceCollection();
-            services.AddHealthChecks()
-                .AddAzureQueueStorage("the-connection-string", queueName: "queue");
+            using var serviceProvider = new ServiceCollection()
+                .AddSingleton(Substitute.For<QueueServiceClient>())
+                .AddHealthChecks()
+                .AddAzureQueueStorage(
+                    (sp, o) => o.QueueName = queueName,
```
sungam3r, line 78: Example of situation when service provider is not needed.

src/HealthChecks.AzureStorage/AzureFileShareHealthCheck.cs
sungam3r: > How do you feel about this instead of the factories? 

Fine.

> Was there a reason I missed for resolving the clients within the method as opposed to the ctor?

I think no. In fact, this is even better, because it allows you to identify configuration errors before actual HC execution.

src/HealthChecks.AzureStorage/AzureFileShareHealthCheck.cs
wsugarman: For the service being unavailable, the message becomes something like: `"Retry failed after 6 tries. Retry settings can be adjusted in ClientOptions.Retry. (No such host is known. (unknown-blob-service.blob.core.windows.net:443)) ..."`. If the service exists but the container does not, we would get a message like `""The specified container does not exist.\nRequestId:5d9aa1a2-d68b-4b31-8f77-9dad95ac0c60\n<a bunch of request metadata>"`.

At the end of the day, the health check endpoint will still just return the phrase "Unhealthy"

src/HealthChecks.AzureStorage/AzureFileShareHealthCheck.cs
sungam3r: > If that's ok, do we need the static client caching?

How will it work without caching?

sungam3r:
I looked through all files so far.

build/dependencies.props
sungam3r: I don't know intents of maintainers of this repo regarding minor/patch but I know that each package can be published independently with its own version.

src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
sungam3r: Indeed. It **can** be superfluous - for **particular** implementation (`DefaultHealthCheckService`). Let's keep try/catch blocks as before. You can post a new issue about removing all try/catch blocks inside HCs across this repo in separate PRif you would like.

src/HealthChecks.AzureStorage/AzureFileShareHealthCheck.cs
sungam3r: ok

src/HealthChecks.AzureStorage/DependencyInjection/AzureStorageHealthCheckBuilderExtensions.cs
```diff
@@ -76,22 +78,94 @@ public static class AzureStorageHealthCheckBuilderExtensions
...
+            this IHealthChecksBuilder builder,
+            Action<AzureBlobStorageHealthCheckOptions>? configureOptions = default,
+            string? name = default,
+            HealthStatus? failureStatus = default,
+            IEnumerable<string>? tags = default,
+            TimeSpan? timeout = default)
+        {
+            var options = new AzureBlobStorageHealthCheckOptions();
```
sungam3r, line 112: see #1356. In general case placing options object outside of `HealthCheckRegistration` and configuring it inside can cause memory leak. Not in this case but nonetheless...

sungam3r: In future one can add some property like `List<>` and add some item inside so then BOOM!

wsugarman: Good catch! I've moved it inside the lambda

src/HealthChecks.AzureStorage/DependencyInjection/AzureStorageHealthCheckBuilderExtensions.cs
wsugarman: Done

src/HealthChecks.AzureStorage/AzureQueueStorageHealthCheck.cs
wsugarman: Restored the outer-try/catch

src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
wsugarman: Restored the outer-try/catch

src/HealthChecks.AzureStorage/AzureFileShareHealthCheck.cs
sungam3r: @wsugarman I'm ready to merge this PR but we have the last unresolved conversation here.

build/dependencies.props
wsugarman: Ah ok. Then probably minor is the most semantically appropriate here

src/HealthChecks.AzureStorage/AzureFileShareHealthCheck.cs
wsugarman: Ah, sorry! I think alternatively we could just create a client in the registration factory and pass it to the ctor. However, the Azure Storage docs do seem to indicate that singleton clients should be used, so caching the client aligns with that guidance. Furthermore, it is already the behavior today, and I think removing it would be too much of a behavioral change.

So we should probably keep the caching haha

wsugarman: @sungam3r I think that's everything

wsugarman:
@sungam3r Thank you so much for your help!

sungam3r:
https://www.nuget.org/packages/AspNetCore.HealthChecks.AzureStorage/6.1.0

src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs
```diff
@@ -7,49 +6,36 @@ namespace HealthChecks.AzureStorage
...
         public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
         {
             try
             {
-                var blobServiceClient = GetBlobServiceClient();
-                await foreach (var page in blobServiceClient.GetBlobContainersAsync(cancellationToken: cancellationToken).AsPages(pageSizeHint: 1))
-                {
-                    break;
-                }
+                await _blobServiceClient.GetPropertiesAsync(cancellationToken);
```
Nisden, line 34: This just broke #783 again

wsugarman: Added new PR here: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/1409

PR #1405, opened 2022-08-14
Allow Use of Table Clients from IServiceProvider

wsugarman:
Adds new health check registration overloads that reuse an existing client from the `IServiceProvider` instead of creating new ones. Includes the following changes:
- Add new registration methods for reusing configured clients
- Add new ctors to each health check that accepts the client
- Use provided cancellation token for underlying API calls
- Standardize the health check logic between the Azure Storage health checks
- Centralize client caching
- Update unit tests

**Which issue(s) this PR fixes**: Relates to https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/1371
Relates to https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/1372

**Does this PR introduce a user-facing change?**: Yes. New APIs

- [X] Code compiles correctly
- [X] Created/updated tests
- [X] Unit tests passing
- [X] Provided sample for the feature

Sample
```csharp
services
    .AddHealthChecks()
    .AddAzureTable(o => o.TableName = "sample-table", name: "sample");
```


test/HealthChecks.CosmosDb.Tests/HealthChecks.CosmosDb.approved.txt
```diff
@@ -3,27 +3,44 @@ namespace HealthChecks.CosmosDb
...
     }
+    public sealed class CosmosDbHealthCheckOptions
+    {
+        public CosmosDbHealthCheckOptions() { }
+        public System.Collections.Generic.IEnumerable<string>? ContainerIds { get; set; }
+        public string? DatabaseId { get; set; }
+    }
     public class TableServiceHealthCheck : Microsoft.Extensions.Diagnostics.HealthChecks.IHealthCheck
     {
-        public TableServiceHealthCheck(string connectionString, string tableName) { }
```
wsugarman, line 14: Following the pattern of the other Azure Storage health checks, I changed the attribute of the `string` to be nullable, so that users may optionally check a given table. This isn't a binary-breaking change, as this is only checked at compile-time, but it may cause code to no longer compile upon update if they use a nullable context.

sungam3r: ok

src/HealthChecks.CosmosDb/TableServiceHealthCheck.cs
```diff
@@ -8,72 +6,51 @@ namespace HealthChecks.CosmosDb
...
                 }
-                var tableClient = tableServiceClient.GetTableClient(_tableName);
-                _ = await tableClient.GetAccessPoliciesAsync();
 
                 return HealthCheckResult.Healthy();
             }
-            catch (RequestFailedException ex) when (ex.Status == 404)
-            {
-                return new HealthCheckResult(context.Registration.FailureStatus, description: $"Table with name {_tableName} does not exist.");
-            }
```
sungam3r, line 63: Why?

src/HealthChecks.CosmosDb/DependencyInjection/CosmosDbHealthCheckBuilderExtensions.cs
```diff
@@ -1,245 +1,394 @@
...
+               failureStatus,
+               tags,
+               timeout));
+        }
+
+        /// <summary>
+        /// Add a health check for Azure CosmosDb/Azure Storage table.
+        /// </summary>
+        /// <remarks>
+        /// A <see cref="TableServiceClient"/> service must be registered in the service container.
```
sungam3r, line 324: I would change 'service container' to 'DI container' everywhere.

src/HealthChecks.CosmosDb/DependencyInjection/CosmosDbHealthCheckBuilderExtensions.cs
```diff
@@ -1,245 +1,394 @@
...
+            return builder.Add(new HealthCheckRegistration(
+               name ?? TABLE_NAME,
+               sp => new TableServiceHealthCheck(endpoint, tokenCredential, tableName),
+               failureStatus,
+               tags,
+               timeout));
+        }
+
+        /// <summary>
+        /// Add a health check for Azure CosmosDb/Azure Storage table.
```
sungam3r, line 321: Azure CosmosDb or Azure Storage  ? Both?

src/HealthChecks.CosmosDb/TableServiceHealthCheck.cs
wsugarman: Similar to the other Azure Storage services, a 404 will also result in an exception whose message will be like `"The table specified does not exist.<more request info>"` so I think special-casing this scenario is superfluous

src/HealthChecks.CosmosDb/DependencyInjection/CosmosDbHealthCheckBuilderExtensions.cs
wsugarman: It's "or" so I can change it

sungam3r:
https://www.nuget.org/packages/AspNetCore.HealthChecks.CosmosDb/6.1.0

PR #1419, opened 2022-08-28
[CosmosDbHealthCheck] Add registration overloads that accept a CosmosClient instance

### LINKED ISSUE #1384: CosmosDbHealthCheck - consider registration overloads that accept a CosmosClient instance
Hi. I appreciate the library and your effort. 👍

I would like to see an overload for CosmosDb health checks that accepts a `CosmosClient` instance.

**Why is this needed**:

According to the official Cosmos SDK v3 documentation (see https://docs.microsoft.com/en-us/azure/cosmos-db/sql/performance-tips-dotnet-sdk-v3-sql#sdk-usage), the `CosmosClient` was designed to be singleton for optimal performance. However, the library does not expose a dependency injection overload that accepts a client, but instead internally creates and maintains a new client for each health check - https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.CosmosDb/CosmosDbHealthCheck.cs#L58.

This can cause a frustrating developer experience when reading cosmos diagnostic logs and seeing more client instances than expected.

danielmpetrov:

Adds new health check registration overloads that allow developers to reuse an existing client(s) instead of creating new ones, without a prerequisite for the clients to be registered in the container.

Resolve #1384 

Yes, new registration overloads for Cosmos DB.

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

Usage

```csharp
var client1 = new CosmosClient("connection string 1");
var client2 = new CosmosClient("connection string 2");

// register other services that use those clients here...

services
    .AddHealthChecks()
    .AddCosmosDb(client)
    .AddCosmosDb(client, options => options.DatabaseId = "db1")
    .AddCosmosDb(client, (services, options) => options.DatabaseId = "db2");
```


danielmpetrov:
I've added the same extensions as introduced in #1405. However, I'm noticing a minor inconvenience when having those two overloads.

It is impossible to not specify the second argument, even though it's optional, as the compiler cannot distinguish the two overloads. E.g.

```csharp
services
    .AddHealthChecks()
    .AddCosmosDb(client); // compiler error here :(
```

The above produces a compiler error, which effectively makes the second argument compulsory, even though it has a default value. This behavior is likely not intentional.

```
Error CS0121 The call is ambiguous between the following methods or properties: 'CosmosDbHealthCheckBuilderExtensions.AddCosmosDb(IHealthChecksBuilder, CosmosClient, Action<CosmosDbHealthCheckOptions>?, string?, HealthStatus?, IEnumerable<string>?, TimeSpan?)' and 'CosmosDbHealthCheckBuilderExtensions.AddCosmosDb(IHealthChecksBuilder, CosmosClient, Action<IServiceProvider, CosmosDbHealthCheckOptions>?, string?, HealthStatus?, IEnumerable<string>?, TimeSpan?)'
```

The below also produces a compiler error (overloads introduced in #1405). The methods in this PR were based on the below.

```csharp
services
    .AddHealthChecks()
    .AddCosmosDb(); // compiler error here :(
```

```
Error CS0121 The call is ambiguous between the following methods or properties: 'CosmosDbHealthCheckBuilderExtensions.AddCosmosDb(IHealthChecksBuilder, Action<CosmosDbHealthCheckOptions>?, string?, HealthStatus?, IEnumerable<string>?, TimeSpan?)' and 'CosmosDbHealthCheckBuilderExtensions.AddCosmosDb(IHealthChecksBuilder, Action<IServiceProvider, CosmosDbHealthCheckOptions>?, string?, HealthStatus?, IEnumerable<string>?, TimeSpan?)'
```

To circumvent this issue, the user would have to do something like the below, which is not pretty.

```csharp
services
    .AddHealthChecks()
    .AddCosmosDb(null as Action<CosmosDbHealthCheckOptions>);
```

sungam3r:
One way is to give different names to methods. Not ideal as well.

test/HealthChecks.CosmosDb.Tests/HealthChecks.CosmosDb.approved.txt
```diff
@@ -41,9 +41,11 @@ namespace Microsoft.Extensions.DependencyInjection
...
         public static Microsoft.Extensions.DependencyInjection.IHealthChecksBuilder AddCosmosDbCollection(this Microsoft.Extensions.DependencyInjection.IHealthChecksBuilder builder, string connectionString, string? database = null, System.Collections.Generic.IEnumerable<string>? collections = null, string? name = null, Microsoft.Extensions.Diagnostics.HealthChecks.HealthStatus? failureStatus = default, System.Collections.Generic.IEnumerable<string>? tags = null, System.TimeSpan? timeout = default) { }
         public static Microsoft.Extensions.DependencyInjection.IHealthChecksBuilder AddCosmosDbCollection(this Microsoft.Extensions.DependencyInjection.IHealthChecksBuilder builder, string accountEndpoint, Azure.Core.TokenCredential tokenCredential, string? database = null, System.Collections.Generic.IEnumerable<string>? collections = null, string? name = null, Microsoft.Extensions.Diagnostics.HealthChecks.HealthStatus? failureStatus = default, System.Collections.Generic.IEnumerable<string>? tags = null, System.TimeSpan? timeout = default) { }
     }
-}
```
sungam3r, line 49: I think you should revert this line to make tests green.

danielmpetrov: I see. It's the `.editorconfig` that forced the newline here https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/cc1f7ceed26d6e229e28f933d60c9b86f3e17b31/.editorconfig#L23.

danielmpetrov:
> One way is to give different names to methods. Not ideal as well.

Another options is to choose one of the overloads, and discard the other. We can keep the more comprehensive one only, i.e. the one with both parameters, and leave up to the users to discard or ignore the first parameter if not needed.

```csharp
services
    .AddHealthChecks()
    .AddCosmosDb(client1) // no overload with one lambda param exists, so no collisions
    .AddCosmosDb(client2, (_, options) => options.DatabaseId = "cosmosdb"); // user can ignore the first parameter if not needed
```

Or else you can decide that the more comprehensive overload that accepts an `IServiceCollection` is not really needed, as the user has direct access to their own services, and leave the simpler overload only.

```csharp
.AddCosmosDb(client2, options => options.DatabaseId = "cosmosdb");
```

danielmpetrov:
> One way is to give different names to methods. Not ideal as well.

We can consider having a single overload instead.

- Either have `Action<IServiceProvider, CosmosDbHealthCheckOptions>`, because it's more comprehensive, and the user can discard or ignore the unused parameter if necessary.

- Or have `Action<CosmosDbHealthCheckOptions>`, if you think the `IServiceProvider` overload does not bring value.

danielmpetrov:
Or else there is a fix to keep the API as it is, and not have it throw a compiler error when the action is omitted. The actions need to be required (we remove the `= default`), and we add a third overload that does not accept a callback action. Then the usage would work.

```csharp
// usage, all three options would be possible then, without any compiler errors
.AddCosmosDb(client)
.AddCosmosDb(client, x => x.DatabaseId = "test")
.AddCosmosDb(client, (x, y) => y.DatabaseId = "test")
```

sungam3r:
50/50 here,  I really don't know.

sungam3r (approved):

danielmpetrov:
Sorry, had to update the approved API order. The tests should pass now. 👍

PR #1473, opened 2022-10-03
Provides ability to use different ctors for AmazonCloudWatchClient

sungam3r:


william-keller:
@sungam3r thanks. this gonna work to me

Issue #1537, opened 2022-11-11
Add Managed Identity support for EventHub in Azure Service Bus


marioleed:
**What would you like to be added**:
TokenCredential (MSI) support for EventHub in the same manner Service Bus (same package) is set up today.

**Why is this needed**:
To be able to use MSI (with all the benefits it entails) instead of a connection string.

marioleed:
I could help out and contribute with a PR.

unaizorrilla:
> I could help out and contribute with a PR.

👌

PR #1540, opened 2022-11-14
Add TokenCredentials for Azure Event Hub

marioleed:
Add TokenCredentials overloads for AddAzureEventHub to simplify the use of MSI. 

Please reference the issue this PR will close: #1537 

You could still use `EventHubConnection` but I'm not sure how well it works with a connection string in its current form. Looking at the code, if the user is using the `AzureEventHubHealthCheck(EventHubConnection connection)` constructor and `AzureEventHubHealthCheck.CheckHealthAsync()` tries to add a new `EventHubProducerClient` (in case the key doesn't exist - which shouldn't happen actually) it will use a connection string without a shared access signature key.

This is a way to get around that.

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing


src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
```diff
@@ -9,55 +10,72 @@ public class AzureEventHubHealthCheck : IHealthCheck
...
 
-            _eventHubConnectionString = $"{connection.FullyQualifiedNamespace};{ENTITY_PATH_SEGMENT}{connection.EventHubName}";
-
-            if (!_eventHubConnections.ContainsKey(_eventHubConnectionString))
-            {
-                _eventHubConnections.TryAdd(_eventHubConnectionString, new EventHubProducerClient(connection));
+            string key = ConnectionKey;
+            if (!_clientConnections.ContainsKey(key))
+            {
+                _clientConnections.TryAdd(key, new EventHubProducerClient(_endpoint, _eventHubName, _tokenCredential));
```
sungam3r, line 59: `TryAdd` _may_ return `false`. I suggest you to look into `internal static class ClientCache` under `HealthChecks.CosmosDb` namespace.

src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
```diff
@@ -9,55 +10,72 @@ public class AzureEventHubHealthCheck : IHealthCheck
...
         public AzureEventHubHealthCheck(EventHubConnection connection)
         {
-            Guard.ThrowIfNull(connection);
+            Guard.ThrowIfNull(connection);
+
+            _connectionString = $"{connection.FullyQualifiedNamespace};{ENTITY_PATH_SEGMENT}{connection.EventHubName}";
+
+            if (!_clientConnections.ContainsKey(_connectionString))
+            {
+                _clientConnections.TryAdd(_connectionString, new EventHubProducerClient(connection));
```
sungam3r, line 46: The same issue with `TryAdd` here.

test/HealthChecks.AzureServiceBus.Tests/DependencyInjection/AzureEventHubUnitWithTokenRegistrationTests.cs
```diff
@@ -0,0 +1,128 @@
...
+            var services = new ServiceCollection();
+            services.AddHealthChecks()
+                .AddAzureEventHub(string.Empty, string.Empty, new AzureCliCredential());
+
+            using var serviceProvider = services.BuildServiceProvider();
+            var options = serviceProvider.GetRequiredService<IOptions<HealthCheckServiceOptions>>();
+
+            var registration = options.Value.Registrations.First();
+
+            Assert.Throws<ArgumentNullException>(() => registration.Factory(serviceProvider));
```
sungam3r, line 125: `Should.Throw<>` + check argument name

marioleed: Should I fix this in the AzureServiceBus* tests as well or should we leave it for later?

src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
marioleed: Added ClientCache 👍🏻

test/HealthChecks.AzureServiceBus.Tests/DependencyInjection/AzureEventHubUnitWithTokenRegistrationTests.cs
sungam3r: Up to you.

src/HealthChecks.AzureServiceBus/ClientCache.cs
```diff
@@ -0,0 +1,28 @@
+using System.Collections.Concurrent;
+
+namespace HealthChecks.AzureServiceBus;
+
+internal static class ClientCache
```
sungam3r, line 5: I suggest to move it into source-shared class (from CosmosDb) + remove another copy from HealthChecks.AzureStorage. Example of source-shared class is `<Compile Include="../CallerArgumentExpressionAttribute.cs" Link="CallerArgumentExpressionAttribute.cs" />` in Directory.Build.props. You may include this file under condition by `MSBuildProjectName` (if you like) to not clutter other projects.

src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
```diff
@@ -1,63 +1,69 @@
...
 
-            _eventHubConnectionString = connectionString.Contains(ENTITY_PATH_SEGMENT) ? connectionString : $"{connectionString};{ENTITY_PATH_SEGMENT}{eventHubName}";
+        _connectionString = $"{connection.FullyQualifiedNamespace};{ENTITY_PATH_SEGMENT}{connection.EventHubName}";
 
-            if (!_eventHubConnections.ContainsKey(_eventHubConnectionString))
-            {
-                _eventHubConnections.TryAdd(_eventHubConnectionString, new EventHubProducerClient(_eventHubConnectionString));
-            }
-        }
+        ClientCache.GetOrAdd(ConnectionKey, _ => new EventHubProducerClient(connection));
```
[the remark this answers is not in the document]
marioleed: Initially I was thinking of scoping this PR to only include EventHub stuff as to not make such a big impact. 

With that said, I was also considering `ClientCache.GetOrAddDisposableAsync` as well. But then thought I wouldn't need it as async methods shouldn't be called from the constructor (which would happen in this case). I guess the problem is the `EventHubConnection` constructor which will not work properly in it's current form with a connection string. I'll try to move it to a field instead. Let's see what I can do 🙂

src/HealthChecks.AzureServiceBus/ClientCache.cs
marioleed: Moved it and scoped it in Directory.Build.props to the three projects using ClientCache, AzureServiceBus, AzureStorage and CosmosDb. However, I have not changed to any async usages in AzureStorage or CosmosDb. Nor have I added ClientCache in the AzureServiceBus* classes. Don't know if I should or not, but I'm fine with that being a new PR 😄

src/HealthChecks.AzureServiceBus/AzureEventHubHealthCheck.cs
```diff
@@ -1,63 +1,73 @@
...
-
-            _eventHubConnectionString = $"{connection.FullyQualifiedNamespace};{ENTITY_PATH_SEGMENT}{connection.EventHubName}";
-
-            if (!_eventHubConnections.ContainsKey(_eventHubConnectionString))
-            {
-                _eventHubConnections.TryAdd(_eventHubConnectionString, new EventHubProducerClient(connection));
-            }
-        }
+            var client = await ClientCache.GetOrAddAsyncDisposableAsync(ConnectionKey,
+                async _ => await Task.FromResult(CreateClient()));
```
sungam3r, line 51: I would not juggle async/await here. Just add another ` public static async ValueTask<T> GetOrAddAsyncDisposableAsync<T>(string key, Func<string, T> clientFactory) where T : IAsyncDisposable` overload and use it here. `IAsyncDisposable` is about asyncronous destroying, not about asyncronous creation.

marioleed: That sounds more sane 😄 Done.

sungam3r (approved):

sungam3r:
Forgot to bump versions in `dependencies.props`.

marioleed:
> Forgot to bump versions in `dependencies.props`.

Ah ok. Yeah that seems interesting. I guess it's mostly minor changes. Otherwise, do advice.

build/dependencies.props
```diff
@@ -14,11 +14,11 @@
     <HealthCheckAWSSqs>6.0.0</HealthCheckAWSSqs>
     <HealthCheckAWSSystemsManager>6.0.0</HealthCheckAWSSystemsManager>
     <HealthCheckAzureDigitalTwin>6.0.0</HealthCheckAzureDigitalTwin>
-    <HealthCheckAzureServiceBus>6.0.4</HealthCheckAzureServiceBus>
-    <HealthCheckAzureStorage>6.1.1</HealthCheckAzureStorage>
+    <HealthCheckAzureServiceBus>6.1.0</HealthCheckAzureServiceBus>
```
marioleed, line 17: @sungam3r Does these bumps look okay?

sungam3r: Yes. Look into approval files. No changes - bump patch. Only green lines - bump minor. Else bump major.

PR #1546, opened 2022-11-25
NATS reconnect

### LINKED ISSUE #1544: NATS check still failing after NATS Server is restarted (AspNetCore.HealthChecks.Nats)
**What happened**:
1. Configure NATS check with NATS Url for readiness:
```
builder.AddNats(options =>
{
    options.Url = "nats://infra-nats:4222";
}, name: "nats", tags: new[]
{
    "ready"
});
```
2. Everyting is working, NATS (deployed in K8s too) is up, healthcheck is working.
3. NATS is down for like 2 minutes due to restart or Kubernetes moving it.
4. Service that has the health check detects it 
5. Service fails readiness probe 
6. NATS is back on-line
7. Service health check still fails despite NATS being accessible again. Readiness probe interval is 15 seconds but we waited like hours and the check never starts working again.
8. Only the service restart helps.

NATS healthcheck detects that NATS is back online.

**Environment**:

- .NET Core version: 
```
.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.11 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.11 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
```
- Healthchecks version:` <PackageReference Update="AspNetCore.HealthChecks.Nats" Version="6.0.3" />`
- Operative system: `alpine 3.16.3` in Docker

sungam3r:
fixes #1544

### DIFF
--- a/src/HealthChecks.Nats/NatsHealthCheck.cs
+++ b/src/HealthChecks.Nats/NatsHealthCheck.cs
@@ -8,8 +8,8 @@ namespace HealthChecks.Nats;
 /// Health check for Nats Server.
 /// </summary>
 /// <remarks>
-/// Relies on a static <see cref="ConnectionFactory"/> which provides factory methods to create connections to NATS Servers,
-/// and a <see cref="NATS.Client.IConnection"/> object connected to the NATS server.
+/// Relies on a static <see cref="ConnectionFactory"/> which provides factory methods to create
+/// connections to NATS Servers, and a <see cref="IConnection"/> object connected to the NATS server.
 /// </remarks>
 public sealed class NatsHealthCheck : IHealthCheck, IDisposable
 {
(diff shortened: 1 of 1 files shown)


src/HealthChecks.Nats/NatsHealthCheck.cs
```diff
@@ -28,9 +28,25 @@ public Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, Canc
...
+                {
+                    connection.Dispose();
+                    connection = exchanged;
+                }
+            }
+
+            // reset connection in case of stuck so the next HC call will establish it again
+            // https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/1544
+            if (connection.State == ConnState.DISCONNECTED || connection.State == ConnState.CLOSED)
+                _connection = null;
```
nosalan, line 47: When this branch is executed (setting `connection` to null) then the code afterwards may throw `System.NullReferenceException` because it's accessing the `State` property in the `GetHealthCheckResultFromState` method.

src/HealthChecks.Nats/NatsHealthCheck.cs
```diff
@@ -28,9 +28,25 @@ public Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, Canc
...
         {
-            _connection ??= CreateConnection(_options);
-            Guard.ThrowIfNull(_connection);
-            var healthCheckResult = GetHealthCheckResultFromState(_connection);
+            // Create new connection if there is no existing one
+            IConnection? connection = _connection;
+            if (connection == null)
+            {
+                connection = CreateConnection(_options);
+                var exchanged = Interlocked.CompareExchange(ref _connection, connection, null);
```
nosalan, line 36: Not sure if I'm aware of the complete picture, but if `CompareExchange` is used to throw away connection objects created by other threads maybe double checked locking would be sufficient on the critical section that creates the connection or using `Lazy<T>`

src/HealthChecks.Nats/NatsHealthCheck.cs
sungam3r: _connection is not connection

src/HealthChecks.Nats/NatsHealthCheck.cs
sungam3r: Eventually they all do the same. I prefer non-blocking APIs if applicable.

PR #1570, opened 2022-12-17
Hide files in Solution Explorer and remove global.json

sungam3r:


sungam3r:
httpbin.org is down now so uris tests failed

PR #1548, opened 2022-11-28
Support current and lts versions

gizmohd:
**What this PR does / why we need it**:  Supports both the Current and LTS versions of dotnet framework 

**Which issue(s) this PR fixes**:  Entity Framework Health Checks Fail at runtime in .net7

Please reference the issue this PR will close: 1533 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:  No

Please make sure you've completed the relevant tasks for this PR, out of the following list:

- [X] Code compiles correctly
- [X] Created/updated tests
- [X] Unit tests passing
- [X] End-to-end tests passing
- [X] Extended the documentation
- [X] Provided sample for the feature

PR #1574, opened 2022-12-17
Prepare projects for v7 packages

sungam3r:
Updates all projects to the latest-and-greatest package versions (v7 series). Replaces #1548 . I found that a lot of 7.x packages from dependencies still support `netstandard2.0` so I switched many existing projects to use it instead of `net6.0`. Some samples were migrated to `net7.0` but only those that have 7.x-dependencies targeted only for `net7.0`.  dependencies.props file was renamed into versions,props since it contains only versions.

`Microsoft.Extensions.Diagnostics.HealthChecks` 7.0.1 still supports netstandard2.0


samples/HealthChecks.UI.Branding/HealthChecks.UI.Branding.csproj
```diff
@@ -1,11 +1,11 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>$(NetFrameworkVersion)</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="6.0.10" />
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="7.0.1" />
```
sungam3r, line 8: NET7-targeted dependency so TFM changed to net7.

sungam3r:
@carlosrecuero @unaizorrilla I am inclined to merge it and move on with other PRs. Please add comments if any.

sungam3r:
Also please pay attention to UI/docker stuff. I don't use either so I don't know all places that should be changed.

PR #1577, opened 2022-12-19
Add Elasticsearch Cluster Health Check

stefannikolei:
This is a second test for elasticsearch. The one implemented now just pings elastic. This one really checks the cluster health. It only respond with unhealthy when the cluster health endpoint answers with red.

Here is the documentation from elastic to this topic:

https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html


src/HealthChecks.Elasticsearch/DependencyInjection/ElasticsearchClusterHealthCheckBuilderExtensions.cs
```diff
@@ -0,0 +1,79 @@
+using HealthChecks.Elasticsearch;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace Microsoft.Extensions.DependencyInjection
```
sungam3r, line 4: Please use file-scoped namespaces for all new code.

src/HealthChecks.Elasticsearch/ElasticsearchClusterHealthCheck.cs
```diff
@@ -0,0 +1,72 @@
...
+
+                    if (!_connections.TryAdd(_options.Uri, lowLevelClient))
+                    {
+                        lowLevelClient = _connections[_options.Uri];
+                    }
+                }
+
+                var healthResponse = await lowLevelClient.Cluster.HealthAsync(ct: cancellationToken);
+                bool isSuccess = healthResponse.ApiCall.HttpStatusCode == 200
+                                 && healthResponse.Status is Health.Green or Health.Yellow;
```
sungam3r, line 60: `Health.Yellow` status may be used to return `HealthCheckResult` in degraded state.

sungam3r (changes requested):
Thanks for contribution. I suggest to implement this feature in the existing healthcheck by providing new property in `ElasticsearchOptions` like `bool UseClusterHealthAPI { get; set; }` - less code.

src/HealthChecks.Elasticsearch/ElasticsearchHealthCheck.cs
```diff
@@ -15,7 +16,8 @@ public ElasticsearchHealthCheck(ElasticsearchOptions options)
             _options = Guard.ThrowIfNull(options);
         }
 
-        public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+        public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context,
+            CancellationToken cancellationToken = default)
```
sungam3r, line 20:
```suggestion
        public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)

```

sungam3r (approved):

PR #1600, opened 2022-12-25
Add ConfigureAwait(false)

sungam3r:
Although for ASP.NET Core applications this is not necessarily (since the synchronization context is null), in the general case `ConfigureAwait` is recommended for library code. Also `ConfigureAwait(false)` may help to avoid deadlocks and speedup tests in Xunit testing environment. See https://github.com/xunit/xunit/blob/main/src/xunit.v3.core/Sdk/MaxConcurrencySyncContext.cs


sungam3r:
@maikebing It looks like CI for InfluxDB hangs, again.

sungam3r:
First run https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/actions/runs/3777596634/jobs/6421511580 - canceled after 32 minutes.

Second run https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/actions/runs/3777624582/jobs/6421553935 - canceled after 20 minutes

Third run OK.

maikebing:
Sometimes it is not possible to return to the download to release information, https://api.github.com/repos/influxdata/influxdb/releases  
`
curl --silent "https://api.github.com/repos/influxdata/influxdb/releases" | jq -r '. | sort_by(.tag_name) | reverse[] | .tag_name' | sed -e 's/v//g' | head -n 1`

Issue #617, opened 2020-09-02
MongoDb health check fails to be healthy when connecting to a database that has no collections yet


cscorley:
Please, fill the following sections to help us fix the issue

**What happened**:

MongoDb health check fails to be healthy when connecting to a database that has no collections.  This is caused by an "Sequence contains no elements" exception.

**What you expected to happen**:

Reports healthy

**How to reproduce it (as minimally and precisely as possible)**:

Create a Mongo database that has no collections.  Specify this database as part of your connection string, e.g.,

mongodb://localhost/test

**Source code sample**:

            services.AddHealthChecks()
                .AddMongoDb(mongodbConnectionString: configuration.GetConnectionString("MongoDB"),
                name: "mongodb",
                failureStatus: HealthStatus.Unhealthy,
                tags: new[] { "ready" });

**Anything else we need to know?**:

We ran into an issue caused by our connection strings with this.  Our setup is to have the client authorize against a particular database which has no collections, then switch to another db.  I understand this is odd and there's a workaround (specify different db in the health check), but the assumption that there will be collections on the db causes a "Sequence contains no elements" exception.

I think the check should use a "FirstOrDefaultAsync" call for both if branches, which I think avoids this situation.  I can also imagine scenarios where new databases are created that do not have collections yet that this scenario would apply.

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/commit/d3a21e2ba2bad10070a2509eba3790c5bc43bdf8#diff-e9b423a67796e60f4cafc9d188b866ee

**Environment**:

- .NET Core version: 3.1
- Healthchecks version: 3.1.2
- Operative system: Windows Server 2016
- Others:

cscorley:
Another alternative to listing databases/collections would be use Ping https://docs.mongodb.com/manual/reference/command/ping/

KasperPurunen:
I have ran into the same issue and have implemented my own version of the health check as follows:
`public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context,
                CancellationToken cancellationToken = default)
            {
                try
                {
                    var mongoClient = MongoClient.GetOrAdd(_mongoClientSettings.ToString(),
                        _ => new MongoClient(_mongoClientSettings));

                    if (string.IsNullOrEmpty(_specifiedDatabase)) return HealthCheckResult.Healthy();

                    using var collectionCursor = await mongoClient
[...]

PR #1626, opened 2023-01-11
MongoDB - Fix Unhealthy if no collections exists in the specified database

ThumNet:
We are experiencing unhealthy checks if a specified mongo database has no collections yet.

```
System.InvalidOperationException: Sequence contains no elements at System.Linq.ThrowHelper.ThrowNoElementsException() at System.Linq.Enumerable.First[TSource](IEnumerable`1 source) at MongoDB.Driver.IAsyncCursorExtensions.FirstAsync[TDocument](IAsyncCursor`1 cursor, CancellationToken cancellationToken) at HealthChecks.MongoDb.MongoDbHealthCheck.CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken) in /_/src/HealthChecks.MongoDb/MongoDbHealthCheck.cs:line 43
```

#617

None

No

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.MongoDb/MongoDbHealthCheck.cs
+++ b/src/HealthChecks.MongoDb/MongoDbHealthCheck.cs
@@ -1,5 +1,6 @@
 using System.Collections.Concurrent;
 using Microsoft.Extensions.Diagnostics.HealthChecks;
+using MongoDB.Bson;
 using MongoDB.Driver;
 
 namespace HealthChecks.MongoDb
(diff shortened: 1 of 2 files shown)


sungam3r (approved):

sungam3r:
Thanks.

sungam3r:
@ThumNet One test failed. I think `mongoClient.GetDatabase(_specifiedDatabase)` returns an instance event for non existing database so `First()` method correctly failed HC before. I'm fine using `FirstOrDefault` but of course some other changes are required to fix failed test.

ThumNet:
> 

Looking into the failing test, will update the PR when fixed.

ThumNet:
After looking into the failing test, which is called `be_unhealthy_on_connectionstring_specified_database_if_mongodb_is_available_and_database_exist`. 

I think that this test should NOT exists, because of the way mongodb works. MongoDB only creates the database when you first store data in that database. This data could be a [collection](https://docs.mongodb.com/manual/reference/glossary/#std-term-collection) or a [document](https://docs.mongodb.com/manual/reference/glossary/#std-term-document).

 Sources: [official mongo docs](hhttps://www.mongodb.com/basics/create-database#option-2) and [blog post](https://theholmesoffice.com/how-to-create-a-mongodb-database/)

So you agree the test can be removed?

sungam3r:
I've read provided links. Do you think the current method of checking the state of MongoDB is sufficient? I have no experience with MongoDB.

ThumNet:
Instead of getting the collections i think it is better to implement a ping to the database. [Others](https://stackoverflow.com/questions/28835833/how-to-check-connection-to-mongodb) also recommend this.

I will change the code to allow this and update the PR. Also change the test accordingly.

ThumNet:
@sungam3r Update has just been pushed.

sungam3r:
@ThumNet Thanks.

Issue #764, opened 2021-04-26
Multiple rabbit MQ reports all connections are unreachable if one connection is down


chevillanos:
Please, fill the following sections to help us fix the issue

**What happened**: Multiple rabbit MQ reports all are unreachable if one connection is down.
Usage: 
```
services.AddHealthChecks()
.AddRabbitMQ(Configuration["rabbitconnStr1"].ToString(), null, "rabbitMqNode1", timeout: TimeSpan.FromSeconds(15))
.AddRabbitMQ(Configuration["rabbitconnStr2"].ToString(), null, "rabbitMqNode2", timeout: TimeSpan.FromSeconds(15));
```

**What you expected to happen**: I am expecting rabbitMqNode1 is healthy if rabbitMqNode2 is down.

**How to reproduce it (as minimally and precisely as possible)**: The usage code above and intentionally stopping the RabbitMq found in rabbitconnStr2 connection while rabbitconnStr1 is up and accessible.
Tested this on a different project while rabbitconnStr1 is reporting as healthy without rabbitconnStr2 declared on the same scenario. Am I chaining them correctly as stated in the readme documentation?

I've also tried having a separate connection with services separately but faces similar behavior:

```
.AddSingleton<IConnectionFactory>(sp=>
            new ConnectionFactory()
            {
                Uri = new Uri(Configuration["rabbitconnStr1"].ToString()),
                AutomaticRecoveryEnabled = true
            })
        .AddHealthChecks()
        .AddRabbitMQ(name: "rabbitMqNode1");

.AddSingleton<IConnectionFactory>(sp=>
            new ConnectionFactory()
            {
                Uri = new Uri(Configuration["rabbitconnStr2"].ToString()),
                AutomaticRecoveryEnabled = true
            })
        .AddHealthChecks()
        .AddRabbitMQ(name: "rabbitMqNode2");
```

*Edit: If I declare rabbitMqNode2 first and rabbitMqNode1 at the bottom shows both endpoints are healthy. I assume whichever comes last overrides the initial declaration? If this is the case, how can I monitor two or more Uri's at the same time?

**Environment**: dev / local

- .NET Core version net 5.0
- Healthchecks version 5.0.1
- Operative system: Windows 10
- Others: Visual Studio 2019 16.9.2

georgy93:
Hello!
[...]

PR #1659, opened 2023-01-28
Allow healthchecking multiple RabbitMQ connections

DPDmancul:
As emerged from #764, it is now impossible to obtain reports from two (or more) RabbitMQ connections. This is due to the fact RabbitMQ connections are singleton instantiated, and so the last added connection overrides the previous ones.

This PR solves this by changing singleton instantiation by lazy initialization. A detailed description is available in a comment on the previous cited issue: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/764#issuecomment-1406389241 

Please reference the issue this PR will close: #764 

Another possibility would be using the multiton pattern. It would be feasible with connections strings, but no so easy with connection factories. 

No: the old behavior of registering one health check for RabbitMQ is preserved. This only allows registering more than one health report.  
If someone already registered more than one health check report for RabbitMQ, he will notice a different behavior. This is not important since he was doing something which was not supported by this library, and indeed he was probably not obtaining what he meant to obtain. In this scenario this PR also solves these misunderstandings.

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [ ] Extended the documentation (not needed*)
- [ ] Provided sample for the feature (not needed)

* Perhaps it would be better to document health checks registers which can be used only once, instead of documenting those which can be added more times. This because it is perfectly valid to call the add extension method more times, and one would normally expect this to add another health check, and not overriding the previous one.

### DIFF
--- a/src/HealthChecks.Rabbitmq/DependencyInjection/RabbitMQHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.Rabbitmq/DependencyInjection/RabbitMQHealthCheckBuilderExtensions.cs
@@ -60,10 +60,9 @@ public static class RabbitMQHealthCheckBuilderExtensions
         IEnumerable<string>? tags = default,
         TimeSpan? timeout = default)
     {
-        var rabbitMQHalthCheck = new Lazy<RabbitMQHealthCheck>(() => new(rabbitConnectionString, sslOption));
         return builder.Add(new HealthCheckRegistration(
             name ?? NAME,
-            _ => rabbitMQHalthCheck.Value,
+            _ => new RabbitMQHealthCheck(rabbitConnectionString, sslOption),
             failureStatus,
             tags,
             timeout));
@@ -73,6 +72,10 @@ public static class RabbitMQHealthCheckBuilderExtensions
     /// Add a health check for RabbitMQ services using <see cref="IConnection"/> from service provider
     /// or <see cref="IConnectionFactory"/> from service provider if none is found. At least one must be configured.
     /// </summary>
+    /// <remarks>
+    /// This method shouldn't be called more than once.
+    /// Each subsequent call will create a new connection, which overrides the previous ones.
+    /// </remarks>
     /// <param name="builder">The <see cref="IHealthChecksBuilder"/>.</param>
     /// <param name="name">The health check name. Optional. If <c>null</c> the type name 'rabbitmq' will be used for the name.</param>
     /// <param name="failureStatus">
@@ -137,24 +140,9 @@ public static class RabbitMQHealthCheckBuilderExtensions
         IEnumerable<string>? tags = default,
         TimeSpan? timeout = default)
     {
-        object rabbitMQHealthCheckLock = new();
-        RabbitMQHealthCheck? rabbitMQHealthCheck = null;
         return builder.Add(new HealthCheckRegistration(
             name ?? NAME,
-            sp =>
-            {
-                if (rabbitMQHealthCheck is null)
-                {
-                    lock (rabbitMQHealthCheckLock)
-                    {
-                        if (rabbitMQHealthCheck is null)
-                        {
-                            rabbitMQHealthCheck = new RabbitMQHealthCheck(connectionFactory(sp));
-                        }
-                    }
-                }
-                return rabbitMQHealthCheck;
-            },
+            sp => new RabbitMQHealthCheck(connectionFactory(sp)),
             failureStatus,
             tags,
             timeout));
@@ -181,24 +169,9 @@ public static class RabbitMQHealthCheckBuilderExtensions
         IEnumerable<string>? tags = default,
         TimeSpan? timeout = default)
     {
-        object rabbitMQHealthCheckLock = new();
-        RabbitMQHealthCheck? rabbitMQHealthCheck = null;
         return builder.Add(new HealthCheckRegistration(
             name ?? NAME,
-            sp =>
-            {
-                if (rabbitMQHealthCheck is null)
-                {
-                    lock (rabbitMQHealthCheckLock)
-                    {
-                        if (rabbitMQHealthCheck is null)
-                        {
-                            rabbitMQHealthCheck = new RabbitMQHealthCheck(connectionFactoryFactory(sp));
-                        }
-                    }
-                }
-                return rabbitMQHealthCheck;
-            },
+            sp => new RabbitMQHealthCheck(connectionFactoryFactory(sp)),
             failureStatus,
             tags,
             timeout));
@@ -227,24 +200,9 @@ public static class RabbitMQHealthCheckBuilderExtensions
         IEnumerable<string>? tags = default,
         TimeSpan? timeout = default)
     {
-        object rabbitMQHealthCheckLock = new();
-        RabbitMQHealthCheck? rabbitMQHealthCheck = null;
         return builder.Add(new HealthCheckRegistration(
             name ?? NAME,
-            sp =>
-            {
-                if (rabbitMQHealthCheck is null)
-                {
-                    lock (rabbitMQHealthCheckLock)
-                    {
-                        if (rabbitMQHealthCheck is null)
-                        {
-                            rabbitMQHealthCheck = new RabbitMQHealthCheck(connectionStringFactory(sp), sslOption);
-                        }
-                    }
-                }
-                return rabbitMQHealthCheck;
-            },
+            sp => new RabbitMQHealthCheck(connectionStringFactory(sp), sslOption),
             failureStatus,
             tags,
             timeout));
--- a/src/HealthChecks.Rabbitmq/DependencyInjection/RabbitMQHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.Rabbitmq/DependencyInjection/RabbitMQHealthCheckBuilderExtensions.cs
@@ -60,12 +60,10 @@ public static class RabbitMQHealthCheckBuilderExtensions
         IEnumerable<string>? tags = default,
         TimeSpan? timeout = default)
     {
-        builder.Services
-            .AddSingleton(sp => new RabbitMQHealthCheck(rabbitConnectionString, sslOption));
-
+        var rabbitMQHalthCheck = new Lazy<RabbitMQHealthCheck>(() => new(rabbitConnectionString, sslOption));
         return builder.Add(new HealthCheckRegistration(
             name ?? NAME,
-            sp => sp.GetRequiredService<RabbitMQHealthCheck>(),
+            _ => rabbitMQHalthCheck.Value,
             failureStatus,
             tags,
             timeout));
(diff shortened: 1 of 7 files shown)


DPDmancul:
You are right, but I didn't find a way to achieve neither of them.

For the second problem I thought a lot about using the lazy pattern also here, but I realized that extension couldn't be called more than once: it would add another identical health check. So maybe it is better to document that it should be called at maximum once, since calling it two times or more is not its objective.

For the first problem one solution could be using the multiton pattern, but we have to check if it is suitable for connection factories. Maybe we can use an unique generated key (if the hash of the factory is not suitable for this purpose). Otherwise we need a way to hook to the disposal of the endpoint, which I don't know if there is a way to.

sungam3r:
> extension couldn't be called more than once: it would add another identical health check.

It could.
```c#
 .ConfigureServices(services =>
                {
                    services
                        .AddSingleton<IConnection>(connection)
                        .AddHealthChecks()
                        .AddRabbitMQ(name: "rmq1", tags: new string[] { "rabbitmq", "module1" })
                        .AddRabbitMQ(name: "rmq2", tags: new string[] { "rabbitmq", "module2" })
                        .AddRabbitMQ(name: "rmq3", tags: new string[] { "rabbitmq", "module3" })
                        .AddRabbitMQ(name: "rmq4", tags: new string[] { "rabbitmq", "module4" })
                        .AddRabbitMQ(name: "rmq5", tags: new string[] { "rabbitmq", "module5" });
                })
```

Ideally all HC extension methods should be idempotent in the sense that with the same input parameters do not lead to duplication of internal objects. Another question is that in practice this will be little in demand, because it is unlikely that the developers will intentionally use the library in the wrong way.

What does _disposal of the endpoint_ mean?

Regarding multiton pattern - I think it's widely used here - search by `ConcurrentDictionary<`. Also I'm prettry sure that each of HC that uses it inside HC class does not call `AddSingleton` at the extension methods level so these HCs support multiple configurations "natively".

DPDmancul:
> Ideally all HC extension methods should be idempotent in the sense that with the same input parameters do not lead to duplication of internal objects. Another question is that in practice this will be little in demand, because it is unlikely that the developers will intentionally use the library in the wrong way.

Adding all those HC has no use other then having the same HC with multiple names. So if we want to manage it we should prevent this misuse, not optimizing it. In my opinion.


> Regarding multiton pattern - I think it's widely used here - search by `ConcurrentDictionary<`. Also I'm prettry sure that each of HC that uses it inside HC class does not call `AddSingleton` at the extension methods level so these HCs support multiple configurations "natively".

Thank you. I think this is the path to follow: I will refactor the PR this way in the next days.

DPDmancul:
Moved to the multiton pattern. I don't understand how this could fix the memory leak problem, since the static dictionary with the connections is not disposed until the class is unloaded. Besides this I think this should be better than the previous lazy pattern, because of being along the lines of other HC.

test/HealthChecks.RabbitMQ.Tests/Functional/RabbitHealthCheckTests.cs
```diff
@@ -238,5 +238,39 @@ public async Task be_healthy_if_rabbitmq_is_available_using_iServiceProvider()
...
+                    app.UseHealthChecks("/health2", new HealthCheckOptions
+                    {
+                        Predicate = r => r.Name.Equals("rabbitmq2")
+                    });
+                });
+
+            using var server = new TestServer(webHostBuilder);
+
+            var response1 = await server.CreateRequest($"/health1").GetAsync().ConfigureAwait(false);
+            var response2 = await server.CreateRequest($"/health2").GetAsync().ConfigureAwait(false);
```
sungam3r, line 270:
```suggestion
            var response1 = await server.CreateRequest("/health1").GetAsync().ConfigureAwait(false);
            var response2 = await server.CreateRequest("/health2").GetAsync().ConfigureAwait(false);
```

src/HealthChecks.Rabbitmq/RabbitMQHealthCheck.cs
```diff
@@ -23,14 +22,17 @@ public RabbitMQHealthCheck(IConnection connection)
...
     }
 
     public RabbitMQHealthCheck(Uri rabbitConnectionString, SslOption? ssl)
     {
-        _rabbitConnectionString = rabbitConnectionString;
-        _sslOption = ssl;
-        _ownsConnection = true;
+        _factory = new ConnectionFactory()
+        {
+            Uri = rabbitConnectionString, AutomaticRecoveryEnabled = true, UseBackgroundThreadsForIO = true,
```
sungam3r, line 31:
```suggestion
            Uri = rabbitConnectionString,
            AutomaticRecoveryEnabled = true,
            UseBackgroundThreadsForIO = true,
```

src/HealthChecks.Rabbitmq/RabbitMQHealthCheck.cs
```diff
@@ -23,14 +22,19 @@ public RabbitMQHealthCheck(IConnection connection)
...
         _factory = Guard.ThrowIfNull(factory);
-        _ownsConnection = true;
     }
 
     public RabbitMQHealthCheck(Uri rabbitConnectionString, SslOption? ssl)
     {
-        _rabbitConnectionString = rabbitConnectionString;
-        _sslOption = ssl;
-        _ownsConnection = true;
+        _factory = new ConnectionFactory()
```
sungam3r, line 29:
```suggestion
        _factory = new ConnectionFactory
```

src/HealthChecks.Rabbitmq/RabbitMQHealthCheck.cs
```diff
@@ -23,14 +22,19 @@ public RabbitMQHealthCheck(IConnection connection)
...
-        _sslOption = ssl;
-        _ownsConnection = true;
+        _factory = new ConnectionFactory
+        {
+            Uri = rabbitConnectionString,
+            AutomaticRecoveryEnabled = true,
+            UseBackgroundThreadsForIO = true,
+        };
+
+        if (ssl!= null)
```
sungam3r, line 36:
```suggestion
        if (ssl != null)
```

src/HealthChecks.Rabbitmq/DependencyInjection/RabbitMQHealthCheckBuilderExtensions.cs
```diff
@@ -139,12 +140,9 @@ public static class RabbitMQHealthCheckBuilderExtensions
         IEnumerable<string>? tags = default,
         TimeSpan? timeout = default)
     {
-        builder.Services
-            .AddSingleton(sp => new RabbitMQHealthCheck(connectionFactory(sp)));
-
         return builder.Add(new HealthCheckRegistration(
             name ?? NAME,
-            sp => sp.GetRequiredService<RabbitMQHealthCheck>(),
+            sp => new RabbitMQHealthCheck(connectionFactory(sp)),
```
sungam3r, line 145: You changed creation semantics here. Before `IConnection` instance was created only once. Now it's created (I understand that user delegate does not necessarily _create_ it but I suppose it's so the majority of time) each call to `HealthCheckRegistration.Factory`. At least I would add xml comment that `connectionFactory` delegate is called each time. The same for other methods accepting delegates.

src/HealthChecks.Rabbitmq/DependencyInjection/RabbitMQHealthCheckBuilderExtensions.cs
```diff
@@ -121,6 +122,10 @@ public static class RabbitMQHealthCheckBuilderExtensions
     /// <summary>
     /// Add a health check for RabbitMQ services using <see cref="IConnection"/> factory function.
     /// </summary>
+    /// <remarks>
+    /// <paramref name="connectionFactory"/> will be called each time the healthcheck route is requested. However the
+    /// created <see cref="IConnection"/> will be reused.
```
sungam3r, line 127: not true

src/HealthChecks.Rabbitmq/RabbitMQHealthCheck.cs
```diff
@@ -49,47 +53,12 @@ public Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, Canc
...
-                    UseBackgroundThreadsForIO = true,
-                };
-
-                if (_sslOption != null)
-                    ((ConnectionFactory)_factory).Ssl = _sslOption;
-            }
-
-            _connection = _factory.CreateConnection();
+            Guard.ThrowIfNull(_factory);
+            _connection = _connections.GetOrAdd(_factory.Uri.ToString(), _ => _factory.CreateConnection());
```
sungam3r, line 59: `SslOption` should be included in key

DPDmancul: This cannot be done for the `RabbitMQHealthCheck(IConnectionFactory)` constructor: `IConnectionFactory` does not expose SSL options (while `ConnectionFactory` does). One solution could be use the hash: `_factory.GetHashCode()` as key.

sungam3r: My concern is about doing something like that:
```c#
.AddRabbitMQ(sp => "amqp://localhost:6672/")
.AddRabbitMQ(sp => "amqp://localhost:6672/", new SslOption { ServerName = "srv", CertPath = "..." })
```

The same uri but different ssl settings. Now the second registration effectively discards ssl options.

DPDmancul: Yes I understood. Even if I don't see any clear use case of doing this, I think we should allow it. For the moment I cannot think nothing better than using the factory itself (`_facotry.GetHashCode()`) as key... but I not really found of it.

sungam3r: Let's take timeout to find solution.

DPDmancul: For RavenDB HC the options object is currently used as key: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/2806c8033352077cda1f2fad5be7dbcf06814637/src/HealthChecks.RavenDB/RavenDBHealthCheck.cs#L28
So using the connection factory as key would be no new concept to this library

sungam3r: So... let's introduce `RabbitMQHealthCheckOptions`. I am inclined to use options in all checks at least for those that I changed.

src/HealthChecks.Rabbitmq/RabbitMQOptions.cs
```diff
@@ -0,0 +1,17 @@
+using RabbitMQ.Client;
+
+namespace HealthChecks.RabbitMQ;
+
+/// <summary>
+/// Options for <see cref="RabbitMQHealthCheck"/>.
+/// </summary>
+public class RabbitMQOptions
```
sungam3r, line 8: `RabbitMQHealthCheckOptions` . All HCs use that naming convention.

src/HealthChecks.Rabbitmq/RabbitMQHealthCheck.cs
```diff
@@ -9,19 +9,22 @@ namespace HealthChecks.RabbitMQ;
...
+    private static readonly ConcurrentDictionary<RabbitMQOptions, IConnection> _connections = new();
 
     private IConnection? _connection;
     private readonly IConnectionFactory? _factory;
+    private readonly RabbitMQOptions _options;
 
     public RabbitMQHealthCheck(IConnection connection)
     {
         _connection = Guard.ThrowIfNull(connection);
+        _options = new RabbitMQOptions(new Uri(connection.Endpoint.ToString()));
```
sungam3r, line 21: When I was talking about options I mean passing options instance into single ctor. Options class contains all needed properties - factory, url, connection, etc. ? See https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.AzureServiceBus/Configuration/AzureEventHubHealthCheckOptions.cs

DPDmancul: The original problem was obtaining SSL config from an IFactory, so it would not be possible to generate a complete RabbitMQHealthCheckConfig from an IFactory. On the other hand we can build an IFactory from an URI and SSL config, so we could use the IFactory instead of RabbitMQHealthCheckConfig.

For what concerns passing a Connection we cannot do much: from it we cannot obtain neither an IFactory, neither a RabbitMQHealthCheckConfig.

sungam3r:
Thanks, @DPDmancul . Superseded by #1910 .

PR #960, opened 2022-02-28
Add ClickHouse DataBase HealthCheck.

Kebastos:
Add ClickHouse DataBase HealthCheck.
Source: [clickhouse.com](https://clickhouse.com/)
This branch is outdated and there are no answers for a long time https://github.com/Xabaril/AspNetCore .Diagnostics.HealthChecks/pull/547

**What this PR does / why we need it**:
Add ClickHouse DataBase HealthCheck because, this branch is outdated and there are no answers for a long time - [547](https://github.com/Xabaril/AspNetCore .Diagnostics.HealthChecks/pull/547)

**Special notes for your reviewer**:
Not all tests pass as it is necessary to configure the ClicHouse database

Please make sure you've completed the relevant tasks for this PR, out of the following list:

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing(pass partially)

PR #1660, opened 2023-01-28
Add ClickHouse DataBase HealthCheck

sungam3r:
Followup of #960


adamsitnik:
The support for ClickHouse was added in #2315, thank you for your contribution @sungam3r !

PR #1661, opened 2023-01-29
Introduce support for timing out redis healthcheck

### LINKED ISSUE #1658: Redis healthchecks don't respect the timeout configure by healthcheck registration
Please, fill the following sections to help us fix the issue

**What happened**:
We've configured the StackExchange Redis library to basically do infinite retries in case of connection errors. 
This means that `ConnectionMultiplexer.ConnectAsync()` will not return til the connection is established. 
This also means that the healthcheck endpoint request never completes till the connection is established. The healthcheck request will not complete in this case till the redis connection is established.

One way of managing this is to use the timeout parameter when setting up the healthcheck registration. Unfortunately the way the redis healthcheck is currently written, it ignores the cancellationtoken passed in.

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/eeb24329d33fb0332ca22ac916516348bd646205/src/HealthChecks.Redis/RedisHealthCheck.cs#L27

`ConnectAsync()` does not take a `cancellationToken` so it would require a bit of extra work checking for the timeout in the library. 

It shouldn't be too complicated to use something like the following to timeout the healthcheck. 

```csharp
var redisConnectionTask = ConnectionMultiplexer.ConnectAsync();
 var timeoutTask = Task.Delay(Timeout.Infinite, cancellationToken);
        var firstTask = await Task.WhenAny(redisConnectionTask, timeoutTask) ;
        if (firstTask == timeoutTask)
        {
            throw new Exception("healthcheck timed out");
        }
```

The healthcheck should fail if the redis connection is not established without the timeout set up as part of the healthcheck registration

Use a connection string like the following for talking to redis
```csharp
  var connectionString =
            "localhost:6379,ssl=False,abortConnect=False,connectRetry=2147483647";
```
I was using redis running in docker. But in this scenario, make sure there is not redis docker container running so there is nothing for the application to connect to

Hit the healthcheck endpoint. Notice the request never completes. 

Try setting a timeout value for the healthcheck as well. It will have no effect. 

**Source code sample**:

**Environment**:

- .NET 6
- Healthchecks version: AspNetCore.HealthChecks.Redis 6.0.4
- Operative system: mac os
- Others:

bdurrani:
Add support for respecting the cancellation token provided by the healthcheck

**Which issue(s) this PR fixes**: #1658 

Please reference the issue this PR will close: #1658 

I only handled the case with a new connection not respecting the healthcheck timeout. 
I'm not sure how I would write an integration test for the case where the connection is made but the rest of the commands time out. So I would prefer to add only the code I can test for. 

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.Redis/RedisHealthCheck.cs
+++ b/src/HealthChecks.Redis/RedisHealthCheck.cs
@@ -81,6 +81,7 @@ public class RedisHealthCheck : IHealthCheck
         }
     }
 
+    // Remove when https://github.com/StackExchange/StackExchange.Redis/issues/1039 is done
     private static async Task<ConnectionMultiplexer> TimeoutAsync(Task<ConnectionMultiplexer> task, CancellationToken cancellationToken)
     {
         using var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
--- a/src/HealthChecks.Redis/RedisHealthCheck.cs
+++ b/src/HealthChecks.Redis/RedisHealthCheck.cs
@@ -24,7 +24,16 @@ public class RedisHealthCheck : IHealthCheck
         {
             if (!_connections.TryGetValue(_redisConnectionString, out var connection))
             {
-                connection = await ConnectionMultiplexer.ConnectAsync(_redisConnectionString).ConfigureAwait(false);
+                var timeoutTask = Task.Delay(Timeout.Infinite, cancellationToken);
+                var connectionMultiplexerTask = ConnectionMultiplexer.ConnectAsync(_redisConnectionString);
+
+                var firstResolved = await Task.WhenAny(timeoutTask, connectionMultiplexerTask).ConfigureAwait(false);
+                if (firstResolved == timeoutTask)
+                {
+                    return new HealthCheckResult(context.Registration.FailureStatus, description: "Healthcheck timed out");
+                }
+
+                connection = await connectionMultiplexerTask.ConfigureAwait(false);
 
                 if (!_connections.TryAdd(_redisConnectionString, connection))
                 {
--- a/src/HealthChecks.Redis/RedisHealthCheck.cs
+++ b/src/HealthChecks.Redis/RedisHealthCheck.cs
@@ -24,16 +24,15 @@ public class RedisHealthCheck : IHealthCheck
         {
             if (!_connections.TryGetValue(_redisConnectionString, out var connection))
             {
-                var timeoutTask = Task.Delay(Timeout.Infinite, cancellationToken);
-                var connectionMultiplexerTask = ConnectionMultiplexer.ConnectAsync(_redisConnectionString);
-
-                var firstResolved = await Task.WhenAny(timeoutTask, connectionMultiplexerTask).ConfigureAwait(false);
-                if (firstResolved == timeoutTask)
+                try
+                {
+                    var connectionMultiplexerTask = ConnectionMultiplexer.ConnectAsync(_redisConnectionString);
+                    connection = await TimeoutAsync(connectionMultiplexerTask, cancellationToken).ConfigureAwait(false);
+                }
+                catch (OperationCanceledException)
                 {
                     return new HealthCheckResult(context.Registration.FailureStatus, description: "Healthcheck timed out");
                 }
-
-                connection = await connectionMultiplexerTask.ConfigureAwait(false);
 
                 if (!_connections.TryAdd(_redisConnectionString, connection))
                 {
@@ -80,5 +79,22 @@ public class RedisHealthCheck : IHealthCheck
             connection?.Dispose();
             return new HealthCheckResult(context.Registration.FailureStatus, exception: ex);
         }
+    }
+
+    private static async Task<ConnectionMultiplexer> TimeoutAsync(Task<ConnectionMultiplexer> task, CancellationToken cancellationToken)
+    {
+        using var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+        var completedTask = await Task.
+            WhenAny(task, Task.Delay(Timeout.Infinite, timeoutCts.Token))
+            .ConfigureAwait(false);
+
+        if (completedTask == task)
+        {
+            timeoutCts.Cancel();
+            return await task.ConfigureAwait(false);
+        }
+
+        cancellationToken.ThrowIfCancellationRequested();
+        throw new OperationCanceledException();
     }
 }
--- a/test/HealthChecks.Redis.Tests/Functional/RedisHealthCheckTests.cs
+++ b/test/HealthChecks.Redis.Tests/Functional/RedisHealthCheckTests.cs
@@ -85,19 +85,19 @@ namespace HealthChecks.Redis.Tests.Functional
         public async Task be_unhealthy_if_redis_is_not_available_within_specified_timeout()
         {
             var webHostBuilder = new WebHostBuilder()
-             .ConfigureServices(services =>
-             {
-                 services.AddHealthChecks()
-                     .AddRedis("nonexistinghost:6379,allowAdmin=true,connectRetry=2147483647", tags: new string[] { "redis" }, timeout: TimeSpan.FromSeconds(2));
+                .ConfigureServices(services =>
+                {
+                    services.AddHealthChecks()
+                        .AddRedis("nonexistinghost:6379,allowAdmin=true,connectRetry=2147483647", tags: new string[] { "redis" }, timeout: TimeSpan.FromSeconds(2));
 
-             })
-             .Configure(app =>
-             {
-                 app.UseHealthChecks("/health", new HealthCheckOptions
-                 {
-                     Predicate = r => r.Tags.Contains("redis")
-                 });
-             });
+                })
+                .Configure(app =>
+                {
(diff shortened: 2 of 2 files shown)


test/HealthChecks.Redis.Tests/Functional/RedisHealthCheckTests.cs
```diff
@@ -80,5 +80,29 @@ public async Task be_unhealthy_if_redis_is_not_available()
...
+                 {
+                     Predicate = r => r.Tags.Contains("redis")
+                 });
+             });
+
+            using var server = new TestServer(webHostBuilder);
+
+            var response = await server.CreateRequest($"/health").GetAsync().ConfigureAwait(false);
+
+            response.StatusCode.ShouldBe(HttpStatusCode.ServiceUnavailable);
```
sungam3r, line 105:
```suggestion
            response.StatusCode.ShouldBe(HttpStatusCode.ServiceUnavailable);
            (await response.Content.ReadAsStringAsync().ConfigureAwait(false)).ShouldContain("Healthcheck timed out");
```

test/HealthChecks.Redis.Tests/Functional/RedisHealthCheckTests.cs
```diff
@@ -80,5 +80,30 @@ public async Task be_unhealthy_if_redis_is_not_available()
...
         }
+
+        [Fact]
+        public async Task be_unhealthy_if_redis_is_not_available_within_specified_timeout()
+        {
+            var webHostBuilder = new WebHostBuilder()
+             .ConfigureServices(services =>
+             {
+                 services.AddHealthChecks()
+                  .AddRedis("nonexistinghost:6379,allowAdmin=true,connectRetry=2147483647", tags: new string[] { "redis" }, timeout: TimeSpan.FromSeconds(2));
```
sungam3r, line 91:
```suggestion
                 services.AddHealthChecks()
                     .AddRedis("nonexistinghost:6379,allowAdmin=true,connectRetry=2147483647", tags: new string[] { "redis" }, timeout: TimeSpan.FromSeconds(2));

```

test/HealthChecks.Redis.Tests/Functional/RedisHealthCheckTests.cs
```diff
@@ -80,5 +80,31 @@ public async Task be_unhealthy_if_redis_is_not_available()
...
+                     .AddRedis("nonexistinghost:6379,allowAdmin=true,connectRetry=2147483647", tags: new string[] { "redis" }, timeout: TimeSpan.FromSeconds(2));
+
+             })
+             .Configure(app =>
+             {
+                 app.UseHealthChecks("/health", new HealthCheckOptions
+                 {
+                     Predicate = r => r.Tags.Contains("redis")
+                 });
+             });
```
sungam3r, line 100:
```suggestion
                .ConfigureServices(services =>
                {
                    services.AddHealthChecks()
                        .AddRedis("nonexistinghost:6379,allowAdmin=true,connectRetry=2147483647", tags: new string[] { "redis" }, timeout: TimeSpan.FromSeconds(2));

                })
                .Configure(app =>
                {
                    app.UseHealthChecks("/health", new HealthCheckOptions
                    {
                        Predicate = r => r.Tags.Contains("redis")
                    });
                });
```

test/HealthChecks.Redis.Tests/Functional/RedisHealthCheckTests.cs
```diff
@@ -80,5 +80,31 @@ public async Task be_unhealthy_if_redis_is_not_available()
...
+
+        [Fact]
+        public async Task be_unhealthy_if_redis_is_not_available_within_specified_timeout()
+        {
+            var webHostBuilder = new WebHostBuilder()
+                .ConfigureServices(services =>
+                {
+                    services.AddHealthChecks()
+                        .AddRedis("nonexistinghost:6379,allowAdmin=true,connectRetry=2147483647", tags: new string[] { "redis" }, timeout: TimeSpan.FromSeconds(2));
+
```
sungam3r, line 92:
```suggestion
```

src/HealthChecks.Redis/RedisHealthCheck.cs
```diff
@@ -24,7 +24,16 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
         {
             if (!_connections.TryGetValue(_redisConnectionString, out var connection))
             {
-                connection = await ConnectionMultiplexer.ConnectAsync(_redisConnectionString).ConfigureAwait(false);
+                var timeoutTask = Task.Delay(Timeout.Infinite, cancellationToken);
```
sungam3r, line 27: Please see https://github.com/graphql-dotnet/server/pull/985 to how to avoid leaking `Task.Delay` instances.

test/HealthChecks.Redis.Tests/Functional/RedisHealthCheckTests.cs
```diff
@@ -80,5 +80,30 @@ public async Task be_unhealthy_if_redis_is_not_available()
...
+                .ConfigureServices(services =>
+                {
+                    services.AddHealthChecks()
+                        .AddRedis("nonexistinghost:6379,allowAdmin=true,connectRetry=2147483647", tags: new string[] { "redis" }, timeout: TimeSpan.FromSeconds(2));
+                })
+                .Configure(app =>
+                {
+                    app.UseHealthChecks("/health", new HealthCheckOptions
+                    {
+                        Predicate = r => r.Tags.Contains("redis")
```
sungam3r, line 97:
```suggestion
                        Predicate = r => r.Tags.Contains("redis"),
                        ResponseWriter = UIResponseWriter.WriteHealthCheckUIResponse,
```

test/HealthChecks.Redis.Tests/Functional/RedisHealthCheckTests.cs
```diff
@@ -80,5 +80,31 @@ public async Task be_unhealthy_if_redis_is_not_available()
...
+                {
+                    services.AddHealthChecks()
+                        .AddRedis("nonexistinghost:6379,allowAdmin=true,connectRetry=2147483647", tags: new string[] { "redis" }, timeout: TimeSpan.FromSeconds(2));
+                })
+                .Configure(app =>
+                {
+                    app.UseHealthChecks("/health", new HealthCheckOptions
+                    {
+                        Predicate = r => r.Tags.Contains("redis"),
+                        ResponseWriter = UIResponseWriter.WriteHealthCheckUIResponse,
```
sungam3r, line 98:
```suggestion
                        ResponseWriter = HealthChecks.UI.Client.UIResponseWriter.WriteHealthCheckUIResponse,
```

src/HealthChecks.Redis/RedisHealthCheck.cs
bdurrani: Thanks for the feedback,
I learned something new today 😄 

I also realized that since we didn't actually cancel the `ConnectionMultiplexer`, it's still trying to constantly reconnect in the background, even though the healthcheck has failed. So every time the healthcheck times out and we hit the healthcheck again, we will keep adding to the `ConnectionMultiplexer`'s attempts to reconnect. I dont know if the `ConnectionMultiplexer` is smart enough to just continue reconnecting again using the last attempt or will it start a whole new set of re-attempts. This might cause problems if people just keep hitting the healthcheck and don't actually restart their applications. 

Ideally if the health check is failing, something bad is happening anyway... so this should be ok?
And once the connection is established, things should resolve themselves.
It's sort of a corner case

src/HealthChecks.Redis/RedisHealthCheck.cs
```diff
@@ -81,4 +80,21 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
...
+            .ConfigureAwait(false);
+
+        if (completedTask == task)
+        {
+            timeoutCts.Cancel();
+            return await task.ConfigureAwait(false);
+        }
+
+        cancellationToken.ThrowIfCancellationRequested();
+        throw new OperationCanceledException();
```
bdurrani, line 98: I still needed a return value here, so I decided to throw the same exception that `ThrowIfCancellationRequested()` would throw.
Technically we should never hit this line because if we're at line 97, that cancellation token has already timed out.

src/HealthChecks.Redis/RedisHealthCheck.cs
sungam3r: I don't know internals of `ConnectionMultiplexer`.

src/HealthChecks.Redis/RedisHealthCheck.cs
```diff
@@ -72,4 +80,21 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
             return new HealthCheckResult(context.Registration.FailureStatus, exception: ex);
         }
     }
+
+    private static async Task<ConnectionMultiplexer> TimeoutAsync(Task<ConnectionMultiplexer> task, CancellationToken cancellationToken)
```
sungam3r, line 84:
```suggestion
    // Remove when https://github.com/StackExchange/StackExchange.Redis/issues/1039 is done
    private static async Task<ConnectionMultiplexer> TimeoutAsync(Task<ConnectionMultiplexer> task, CancellationToken cancellationToken)
```

PR #1713, opened 2023-02-27
Healthcheck checks all hosts if multiple added with AddHost

### LINKED ISSUE #1710: Health check for multiple hosts should not return on the first failing host
My suggestion is that health check should not return on the first failing host, but try all other hosts registered with AddHost and return a message with all failing hosts. FolderHeathCheck returns a error description for each missing folders

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.System/FolderHealthCheck.cs#LL26C9-L26C9

Most of the other HealthCheck e.g.

- Ping https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.Network/PingHealthCheck.cs#LL28C28-L28C28
- TCP https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.Network/TcpHealthCheck.cs#L32
- SSL https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.Network/SslHealthCheck.cs#L42

return the erro on the first failing host. Therefore if multiple hosts fail than only one is reported. It would be nice if all healthchecks that have the AddHost method to add multiple hosts support the error message construction like FolderHeathCheck does.

For now the workaround is to use multple time AddPingHealthCheck, AddTcpHealthCheck, ...

michaelmairegger:
Modify some health checks in such a way that they will not return failure state on first failing host but continue with other hosts. This allows to view the state of all hosts, if others are failing

This is a preview of what I wanted to point out with the opened issue. If this work is OK, then I will continue for other health-checks and add further commits to this PR

Closes #1710


sungam3r:
I think that for all checks that check several entities (folders, hosts, drives, etc), it makes sense to add an option like `CheckAllDrives` by analogy with one from `DiskStorageOptions`.

michaelmairegger:
@sungam3r  done

sungam3r:
Conflicts here.

> If you want I can continue for other checks too?

OK.

src/HealthCheckResultTask.cs
```diff
@@ -10,3 +10,27 @@ internal static class HealthCheckResultTask
     /// </summary>
     public static readonly Task<HealthCheckResult> Healthy = Task.FromResult(HealthCheckResult.Healthy());
 }
+
```
sungam3r, line 13:
```suggestion
```

src/HealthCheckResultTask.cs
```diff
@@ -10,3 +10,26 @@ internal static class HealthCheckResultTask
...
+    public Task<HealthCheckResult> GetHealthStateAsync(HealthCheckContext context)
+    {
+        if (Count > 0)
+        {
+            return Task.FromResult(new HealthCheckResult(context.Registration.FailureStatus, description: string.Join("; ", this)));
+        }
+
+        return HealthCheckResultTask.Healthy;
+    }
+
```
sungam3r, line 25:
```suggestion
```

sungam3r: Also I would switch to extension method for `List<string>` so insteaf of `return errorList?.GetHealthState(context) ?? HealthCheckResult.Healthy();` you can write `return errorList.GetHealthState(context);`

src/HealthCheckResultTask.cs
```diff
@@ -10,3 +10,15 @@ internal static class HealthCheckResultTask
     /// </summary>
     public static readonly Task<HealthCheckResult> Healthy = Task.FromResult(HealthCheckResult.Healthy());
 }
+
+internal static class StringListExtensions
+{
+    public static HealthCheckResult GetHealthState(this List<string> instance, HealthCheckContext context)
+    {
+        if (instance.Count == 0)
```
sungam3r, line 18:
```suggestion
    public static HealthCheckResult GetHealthState(this List<string>? instance, HealthCheckContext context)
    {
        if (instance?.Count ?? 0 == 0)
```

src/HealthChecks.Network/DnsResolveHealthCheck.cs
```diff
@@ -20,6 +20,7 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
     {
         try
         {
+            List<string> errorList = new();
```
sungam3r, line 23:
```suggestion
            List<string>? errorList = null;

```

src/HealthChecks.Network/DnsResolveHealthCheck.cs
```diff
@@ -32,12 +33,16 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
                 {
                     if (item.Resolutions == null || !item.Resolutions.Contains(ipAddress.ToString()))
                     {
-                        return new HealthCheckResult(context.Registration.FailureStatus, description: $"Ip Address {ipAddress} was not resolved from host {item.Host}");
+                        errorList.Add($"Ip Address {ipAddress} was not resolved from host {item.Host}");
```
sungam3r, line 36:
```suggestion
                        (errorList ??= new()).Add($"Ip Address {ipAddress} was not resolved from host {item.Host}");

```

sungam3r: I would avoid memory allocations for green path.

michaelmairegger: Do you mean, that errorList should only be initialized if there is an error?

src/HealthCheckResultTask.cs
michaelmairegger: Doesn't it look strange if the GetHealthState method works on `null` objects? Writing the suggested code will allow the following code:

```
List<string>? t = null;
if (t is null)
{
    t.GetHealthState(context);
}
```

Code is valid and works but this is weird.

src/HealthChecks.Network/DnsResolveHealthCheck.cs
michaelmairegger:
```
public static class EmptyList<T>
{
    internal static readonly List<T> Value = new List<T>();
}
```

And initialize it with

`List<string> errorList = EmptyList<string>.Value;`

With this helper class no additional memory is allocated.

sungam3r: > Do you mean, that errorList should only be initialized if there is an error?

Yes.

> With this helper class no additional memory is allocated.

It does not work. Static instance will be used by all checks concurrently.

src/HealthCheckResultTask.cs
sungam3r: > Doesn't it look strange if the GetHealthState method works on null objects? 

In our case - no, because `null` means the _no errors_.

michaelmairegger:
Don't know what those failing tests mean.

sungam3r:
They mean that your PR changed public API surface. Each test project has *.approved.txt file that tracks public APIs. Open your IDE for tests, navigate to particular `ApiApprovalTests` test and run it. It will generate `*.received.txt` file in case of changes. Delete approved (old one) and rename received to approved.

sungam3r:
Thanks.

PR #1717, opened 2023-02-27
Fixes some issue with docker-compose.yml

michaelmairegger:
This PR fixes the following issues regarding docker-compose.yml

- [x]  Used image is not available anymore
- [x]  Start `healthchecksui` when `sqlserver` is up and running


sungam3r:
I've never used docker but... looks good.

PR #1581, opened 2022-12-20
Add functionality to hide exceptions from health check response

worldpwn:
**What this PR does / why we need it**:
It is useful to have the option to remove exception messages for the production environment.

**Which issue(s) this PR fixes**:

Please reference the issue this PR will close: #_[issue number]_

**Special notes for your reviewer**:
Usage:

```csharp
app
    .UseRouting()
    .UseEndpoints(config =>
    {
        config.MapHealthChecks("/healthz", new HealthCheckOptions
        {
            Predicate = _ => true,
            ResponseWriter = UIResponseWriter.WriteHealthCheckUIResponseNoExceptions
        });
    });
```

**Does this PR introduce a user-facing change?**:

Please make sure you've completed the relevant tasks for this PR, out of the following list:

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [x] End-to-end tests passing
- [ ] Extended the documentation
- [ ] Provided sample for the feature

PR #1719, opened 2023-03-01
Feature to return AzureServiceBus Queue health exception description based on boolean input parameter.

### LINKED ISSUE #1693: Add control over Health Check response
**A feature to ignore Exception or Description from the response when .AddAzureServiceBusQueue(connectionstring, queuename, name) is being added to ServiceCollection for HealthChecks.**

**With the response of type HealthCheckResult , I do not want to expose in detail exception and description of Unhealthy Queues. So a feature allowing control on the response will help to control what to expose and what not to.**:

akm151:
**This PR adds a feature to the AzureServiceBus Queue HealthCheck. It simply gives you a boolean parameter to set if you require the Queue failure description and exception in the situation your queues are unhealthy.**:
```C#
services
   . AddHealthChecks()
   . AddAzureServiceBusQueueMessageCountThreshold(connectionString: "", queueName: "", isExceptionDetailsRequired: false);
```
Closes #1693 

**Does this PR introduce a user-facing change?
This PR adds a feature for users relying on AzureServiceBusQueue HealthCheck**:

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing


src/HealthChecks.AzureServiceBus/HealthChecks.AzureServiceBus.csproj
```diff
@@ -5,6 +5,7 @@
     <PackageTags>$(PackageTags);Azure;ServiceBus;EventHub</PackageTags>
     <Description>HealthChecks.AzureServiceBus is the health check package for Azure EventHub, and Azure Service Bus Queues and Topics.</Description>
     <VersionPrefix>$(HealthCheckAzureServiceBus)</VersionPrefix>
+    <Version>7.0.3</Version>
```
sungam3r, line 8:
```suggestion
```

src/HealthChecks.AzureServiceBus/DependencyInjection/AzureServiceBusHealthCheckBuilderExtensions.cs
```diff
@@ -244,15 +245,17 @@ public static class AzureServiceBusHealthCheckBuilderExtensions
         string? name = default,
         HealthStatus? failureStatus = default,
         IEnumerable<string>? tags = default,
-        TimeSpan? timeout = default) =>
+        TimeSpan? timeout = default,
+        bool isExceptionDetailsRequired = true) =>
```
sungam3r, line 249: I would not touch any of extension methods here to add one more argument. The appointment of an option class (for any health check) is to facilitate the support and configuration of health checks. This method (as well as some others) contains `Action<AzureServiceBusQueueHealthCheckOptions>? configure` delegate, so just set the value of desired setting inside it.

src/HealthChecks.AzureServiceBus/Configuration/AzureServiceBusQueueHealthCheckOptions.cs
```diff
@@ -20,6 +20,7 @@ public class AzureServiceBusQueueHealthCheckOptions : AzureServiceBusHealthCheck
     /// is used set this to <see langword="false"/>. By default <see langword="true"/>.
     /// </remarks>
     public bool UsePeekMode { get; set; } = true;
+    public bool IsExceptionDetailsRequired { get; set; } = true;
```
sungam3r, line 23:
```suggestion
    public bool CaptureException { get; set; } = true;
```

It seems that such option is applicable to every health check into solution.

sungam3r: After thinking about that a bit I'm sure this option is useless.
1. This behaviour is not specific to that HC. Such option is applicable to every HC.
2. By default ASP.NET Core captures exception but it writes plain text response with only status code, see ` public Func<HttpContext, HealthReport, Task> ResponseWriter { get; set; } = HealthCheckResponseWriters.WriteMinimalPlaintext;
`
3. One can set custom `ResponseWriter`. Indeed this repo does, see `UIResponseWriter.WriteHealthCheckUIResponse`.
4. Also #1581 provides a feture to hide exception text.
5. You may filter exceptions in ` public static UIHealthReport UIHealthReport.CreateFrom(HealthReport report, Func<Exception, string>? exceptionMessage = null)`.

akm151: @sungam3r  is https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/1581 present in version 6 of the package?

akm151: Closing PR. Will wait for the feature  in 7.x.x release.

Issue #1738, opened 2023-03-11
SMTP Health check does not respect timeout


peterwurzinger:
Please, fill the following sections to help us fix the issue

**What happened**:
When configuring a timeout for a SMTP check, it is only respected for SMTP commands, but not for the TCP connection itself. When the host is not reachable, it could be the case that the given timeout is exceeded by far.

**What you expected to happen**:
The timeout set is respected even for the initial TCP handshake.
**Note:** This can cause your Kubernetes pods (if using in this environment) to be considered as down, if the timeout for the TCP handshake exceeds the healthcheck timeout configured in your gateway/ingress.

**How to reproduce it (as minimally and precisely as possible)**:
```cs
healthCheckBuilder.AddSmtpHealthCheck(options => {
  options.Host = "anyInvalidHost";
}, timeout: TimeSpan.FromSeconds(5));
```

**Anything else we need to know?**:
I already opened a PR that should fix this issue:
#1739

**Environment**:

- .NET Core version: 6
- Healthchecks version: 6.0.5
- Operative system: Windows 10 and Ubuntu (Kubernetes)

peterwurzinger:
Fixed by #1739

PR #1739, opened 2023-03-11
Fix SMTP health checks not respecting timeout

peterwurzinger:
The cancellation token in an SMTP handshake will be passed to the initial TCP connection as well, making it respect a given timeout.
It seems like this has already been an issue for IMAP connections, as there is a [Test](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/a3c8294bd4edbd1658152971fd8aac4f12e4f01f/test/HealthChecks.Network.Tests/Functional/ImapHealthCheckTests.cs#L280) and the same [Fix](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/a3c8294bd4edbd1658152971fd8aac4f12e4f01f/src/HealthChecks.Network/ImapHealthCheck.cs#L30) for that type of checks.

#1738

**Does this PR introduce a user-facing change?**
At least I'm not aware of any.

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.Network/Core/MailConnection.cs
+++ b/src/HealthChecks.Network/Core/MailConnection.cs
@@ -3,6 +3,9 @@ using System.Net.Security;
 using System.Net.Sockets;
 using System.Security.Cryptography.X509Certificates;
 using System.Text;
+#if !NET5_0_OR_GREATER
+using HealthChecks.Network.Extensions;
+#endif
 
 namespace HealthChecks.Network.Core;
 
@@ -28,10 +31,14 @@ public class MailConnection : IDisposable
         _allowInvalidCertificates = allowInvalidCertificates;
     }
 
-    public async Task<bool> ConnectAsync()
+    public async Task<bool> ConnectAsync(CancellationToken cancellationToken = default)
     {
         _tcpClient = new TcpClient();
-        await _tcpClient.ConnectAsync(Host, Port).ConfigureAwait(false);
+#if NET5_0_OR_GREATER
+        await _tcpClient.ConnectAsync(Host, Port, cancellationToken).ConfigureAwait(false);
+#else
+        await _tcpClient.ConnectAsync(Host, Port).WithCancellationTokenAsync(cancellationToken).ConfigureAwait(false);
+#endif
 
         _stream = GetStream();
         await ExecuteCommand(string.Empty).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


sungam3r:
Thanks. Please look into #1740 as an example of how this can be done for modern TFMs. Let's pass `cancellationToken` down the call tree and use overloads that accept it. Use `WithCancellationTokenAsync` only as a fallback for other TFMs.

peterwurzinger:
Alright, i finally had time for that. Please have a look if that is how you wanted it to be implemented.

But allow me a few additional notes:
1. I had to leverage the `CancellationToken` all the way up to `MailConnection.ConnectAsync` as a parameter. That is not necessarily a problem, because there is then only one place to do this TFM-dance. BUT: `ImapConnection` is also derives from `MailConnection`, yet I did not touch its implementation within this PR to not mix things up.  Imo it would absolutely make sense to pass the `CancellationToken` in IMAP connections as well.
2. It now is a binary breaking change. Since `ConnectAsync` accepts a `CancellationToken` as default parameter, it doesn't break builds, yet if only the binary would be replaced during runtime, one would see runtime errors. (Don't know if somebody would really do that and use the "raw" MailConnection , but yeah...)
3. There are quite a few places in the mail-specific check implementations, where a Cancellation token could be passed, or an async alternative could be used. e.g.
  - AuthenticateAsync
  - ExecuteCommand (yet I don't know if that is maybe intentional to keep commands as atomic operations)
  - AuthenticateAsClientAsync

Maybe you should share your thoughts on that.

sungam3r:
> It now is a binary breaking change

It doesn't matter for now since all packages are in 7.0.0-preview state.

peterwurzinger:
Fair enough, yes.

sungam3r:
@peterwurzinger Finally I returned to review this one.

> 
    AuthenticateAsync
    ExecuteCommand (yet I don't know if that is maybe intentional to keep commands as atomic operations)
    AuthenticateAsClientAsync


Yes, I'm fine having them all with `Async` suffix and accepting `CancellationToken`. Also place ` ArrayPool<byte>.Shared.Return(readBuffer);` under `finally` in `ExecuteCommand`.

peterwurzinger:
Shall this be done within this PR, or shall I open a separate one? I guess it would touch not only the SMTP implementation, but also the IMAP check, therefore being probably a lot more than just fixing #1738

peterwurzinger:
@sungam3r  I opened another PR to cover those additional chores:
https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/1812

PR #1740, opened 2023-03-13
Add missed net5/net6 TFMs

sungam3r:
These projects contain ifdefs.


michaelmairegger:
net5.0 is EOL since May 2022
https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

sungam3r:
It does not matter much for library packages that are intended to be consumed by someone. TFM is just a identifier that describes minimum required API set from consumer. So net5-targeted library works normally in net6/net7 and so on app.

Issue #1714, opened 2023-02-27
Looking for maintainers


sungam3r:
Hi there. For a long time I have been helping to support this project. At first I only used some health checks from it, then I began to offer small changes. Over time, my knowledge and experience became enough for the core maintainers could calmly entrust me with the management of the project and give the necessary permissions to independently change anything in it. I helped and continue to help. Nevertheless, this is not the main project that I am taking time. My knowledge and time is not enough to carry out decent support here by oneself. Many issues are unanswered. The same thing, although to a lesser extent concerns pull requests. 

Now I'm the only person who sorts issues, gives any feedback in reviewing PRs, merges PRs, makes releases, etc. I have not received feedback and have not seen any activity from core maintainers - @CarlosLanderas @carlosrecuero @unaizorrilla for that repo for a very long time.

So... I would be very glad if someone is ready to help us to support this project **on an ongoing basis**. The main pain points so far are
- UI projects
- ~CI for UI does not call dotnet test~ DONE
- Docker/K8S stuff - it all needs review
- ~KubernetesClient nuget package is very old - v4.0.26 (09.04.2021) vs 10.0.1 on nuget.org~ DONE
- A lot of pending/unanswered issues/PRs.

This project is worthy of continuing to live.

marioleed:
@sungam3r It seems to me that you're practically the only maintainer, which feels a bit odd for a popular library like this.  You are doing a great job, so keep it up 💪🏻

I don't know if I can be of any help, but I can look through some PR's and issues when I have some spare time (don't hold your breath waiting though 😉).

sungam3r:
> which feels a bit odd for a popular library like this

In fact, this can be seen everywhere in OSS. Only 1-2 people on an ongoing basis support projects with millions of downloads.

rob-baldwin:
@sungam3r I'm willing to put my hand up to help with the long term support of this project.  I've not contributed much to this project, but I've used it a fair bit in several companies over many years, so I feel I should give back to the community.

sungam3r:
OK. You can start by answering issues marked with `question` label or/and post some PRs. I'll review them as soon as possible. Also you may find a buch of abandoned PRs with no feedback (#1121, for example) from OP, so another way to help - resurrect such PRs so we can merge them.

SeanFarrow:
@sungam3r I'm also interested in helping.
[...]

PR #1782, opened 2023-04-02
Fixes #716 ui settings api query failing.

### LINKED ISSUE #716: HealthChecksUI query api endpoint settings fails when using ApplicationBuilder setup
**What happened**: Setup using  the ApplicationBuilder  pattern `app.UseHealthChecksUI();`, causes the HealthChecksUI  to display partial information and no updates. There is no title and the textbox that display the polling interval is blank.

**What you expected to happen**: Separate setups for either`app.UseHealthChecksUI();` or`endpoints.MapHealthChecksUI();`  should work and produce a configuration that allows the UI to be displayed and be updated based on the pollingInterval. 

Swap the commenting of  `app.UseHealthChecksUI();` [ Fails] and `endpoints.MapHealthChecksUI(); ` [Works]
between execution attempts to see the behavior.

```
 public void ConfigureServices(IServiceCollection services)
{
    services.AddControllers();
    services.AddHealthChecks();
    services.AddHealthChecksUI().AddInMemoryStorage();
}

// This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
{
    // Fails, healthchecks-api/ui-settings returns the wrong object. 
    // Appears to be test results when fully configured with actual checks.
    // []
    // and the SPA is not populated with needed title or pollingInterval properties.
        app.UseHealthChecksUI();

    app.UseRouting();

    app.UseEndpoints(endpoints =>
    {
        endpoints.MapControllers();
        // Works as expected, healthchecks-api/ui-settings returns default/appsettings values.
        // {"pollingInterval":10,"headerText":"Health Checks Status"}
        //endpoints.MapHealthChecksUI();  
    });
}
```

Also calling the api directly `healthchecks-api/ui-settings` will return different objects.  When using the setup with MapHealthChecksUI the api returns what appears to be proper JSON settings ` {"pollingInterval":10,"headerText":"Health Checks Status"}` . While using the setup with ` app.UseHealthChecksUI`  the api returns something different.

**Source code sample**:

**Anything else we need to know?**:  I walked things back to the Keys.cs file. Interesting that two separate constants are assigned the same value. I wonder if some confusion about which is which may have occurred.  I stopped there.

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.UI/Keys.cs
```
namespace HealthChecks.UI
{
    internal class Keys
    {
	internal const string HEALTHCHECKS_SETTINGS_ENDPOINT = "ui-settings";
         .
         .
         .
 	internal const string HEALTHCHECKSUI_SETTINGS_PATH = "ui-settings";
```
**Environment**:

- .NET Core version: 5.0.2
- Healthchecks version: 5.0.1
- Operative system: Windows 10
- Others:

real-jrw2:
When using the ApplicationBuilder.UseHealthChecksUI, the ui settings middleware is never executed.  The ui api endpoint middleware is currently handling all requests to "healthchecks-api/*".  Moving the mapping of the ui settings middleware to before the ui api endpoint middleware mapping allows the ui settings to handle requests to "healthchecks-api/ui-settings".

This does put the ui-settings behind the request limiter, if that was not the intention then the ui settings mapping should happen before the api path is mapped.
716

no

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/src/HealthChecks.UI/Extensions/ApplicationBuilderExtensions.cs
+++ b/src/HealthChecks.UI/Extensions/ApplicationBuilderExtensions.cs
@@ -28,12 +28,11 @@ namespace Microsoft.AspNetCore.Builder
             {
                 appBuilder
                 .UseMiddleware<UIApiRequestLimitingMidleware>()
+                .Map($"/{Keys.HEALTHCHECKSUI_SETTINGS_PATH}", appBuilder => appBuilder.UseMiddleware<UISettingsMiddleware>())
                 .UseMiddleware<UIApiEndpointMiddleware>();
             });
 
             app.Map(options.WebhookPath, appBuilder => appBuilder.UseMiddleware<UIWebHooksApiMiddleware>());
-
-            app.Map($"{options.ApiPath}/{Keys.HEALTHCHECKSUI_SETTINGS_PATH}", appBuilder => appBuilder.UseMiddleware<UISettingsMiddleware>());
 
             new UIResourcesMapper(
                 new UIEmbeddedResourcesReader(embeddedResourcesAssembly))


sungam3r:
@real-jrw2 Thanks for contribution. Unfortunately, now this project has little support, especially in UI-related stuff, see #1714. Also CI workflow for UI projects is broken so I tend not to merge any PRs until it will be fixed.

sungam3r:
UI CI was fixed, conflicts here

Issue #1798, opened 2023-04-16
Move from Newtonsoft.Json to System.Text.Json


preardon:
Remove Newtonsoft as a dependency and user the newer faster System.Text

Benefits would be Performance and less dependant packages./

I would be happy to do the work

sungam3r:
PR is welcome. See #1712 for run UI tests on CI. These tests use serialization.

sungam3r:
UI CI was fixed.

preardon:
Excellent, initially I might love the tests with Newtonsoft to help ensure no regression

sungam3r:
@preardon This will be published in v7.

PR #1869, opened 2023-07-05
Switch from Newtonsoft.Json to System.Text.Json

sungam3r:
Rel: #1798


src/HealthChecks.Publisher.Seq/SeqPublisher.cs
```diff
@@ -51,15 +51,15 @@ public async Task PublishAsync(HealthReport report, CancellationToken cancellati
...
+                        { "RawReport" , JsonSerializer.Serialize(report)}
                     }
                 }
             }
         };
 
         _options.Configure?.Invoke(events);
 
-        await PushMetricsAsync(JsonConvert.SerializeObject(events), cancellationToken).ConfigureAwait(false);
+        await PushMetricsAsync(JsonSerializer.Serialize(events), cancellationToken).ConfigureAwait(false);
```
sungam3r, line 62: May require further customization in future.

sungam3r:
I guess I could miss to add some converters in STJ settings for Seq publisher to work properly but we may fix it later with patch release.

sungam3r:
Waiting for comments 1-2 days.

PR #1898, opened 2023-07-11
Fixed docker compose (consul service)

turric4n:

Fix docker-compose up

Any

Any

- [V ] Code compiles correctly
- [V] Unit tests passing
- [V] End-to-end tests passing

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -134,7 +134,7 @@ services:
         - "${TMPDIR:-/tmp/localstack}:/tmp/localstack"
         - "/var/run/docker.sock:/var/run/docker.sock"
     consul:
-        image: consul:1.15.4
+        image: hashicorp/consul:latest
         ports:
           - "8500:8500"
           - "8600:8600"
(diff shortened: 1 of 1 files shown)


sungam3r:
What problem does this PR fix?

sungam3r:
I'm not an expert in docker and I've never used anything related to docker from that repo. I don't know the purpose of that file.

sungam3r:
Also there is no CI run triggered.

turric4n:
Current docker-compose file doesn't work because consul latest image not exists. I fixed that. That should be merged asap. Thanks.

turric4n:
> I'm not an expert in docker and I've never used anything related to docker from that repo. I don't know the purpose of that file.

That's ok. A Docker Compose file simplifies the orchestration of Docker services. 

Here we use Docker Compose to easily set up services so that tests can utilize these services locally.

docker-compose.yml
```diff
@@ -134,7 +134,7 @@ services:
         - "${TMPDIR:-/tmp/localstack}:/tmp/localstack"
         - "/var/run/docker.sock:/var/run/docker.sock"
     consul:
-        image: consul:latest
+        image: consul:1.15.4
```
rob-baldwin, line 137:
```suggestion
        image: hashicorp/consul:latest
```

We should be using the updated supported image, not the deprecated image.

turric4n: You are right! Thanks!

sungam3r: Ah, I fixed that in #1853 for consul CI. OK.

PR #1952, opened 2023-07-24
Fixes some build issues

michaelmairegger:
Fixes some build issues

### DIFF
--- a/src/HealthChecks.UI.Data/HealthChecksDb.cs
+++ b/src/HealthChecks.UI.Data/HealthChecksDb.cs
@@ -5,23 +5,21 @@ namespace HealthChecks.UI.Data
 {
     public class HealthChecksDb : DbContext
     {
-        public DbSet<HealthCheckConfiguration> Configurations { get; set; }
+        public DbSet<HealthCheckConfiguration> Configurations => Set<HealthCheckConfiguration>();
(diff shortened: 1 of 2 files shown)


src/HealthChecks.UI.Data/HealthChecksDb.cs
```diff
@@ -5,23 +5,21 @@ namespace HealthChecks.UI.Data
 {
     public class HealthChecksDb : DbContext
     {
-        public DbSet<HealthCheckConfiguration> Configurations { get; set; }
+        public DbSet<HealthCheckConfiguration> Configurations => Set<HealthCheckConfiguration>();
```
sungam3r, line 8: Why is this done?

michaelmairegger:
> What issues?

I have removed the pragma warning for CS8618 and made use of Set<T> to prevent CS8618 according https://learn.microsoft.com/en-us/ef/core/miscellaneous/nullable-reference-types#dbcontext-and-dbset

src/HealthChecks.UI.Data/HealthChecksDb.cs
michaelmairegger: Never mind, false positive. I will close the PR

PR #1981, opened 2023-08-04
OpenID Connect - Validate Discovery Endpoint response

### LINKED ISSUE #701: OpenID Connect server looks only at HTTP response code
**What happened**:

I performed a test health check to Identity Server and, to check its behaviour, I had an incorrect URL on purpose: "/ssox/" instead of "/sso/", where "/sso/" is reverse proxied to Identity Server. On "/" and below, all non-SSO requests return a single-page web application with 200 OK, so this occurred for "/ssox/" as well. There is no possibility for me to change this as the application depends on it.

To my surprise, the health check succeeded. This is because the health check only looks at HTTP status code, not whether the response makes any sense otherwise. However, from the HTTP response it is clearly visible that it is not Identity Server responding:

- content-type is not application/json but text/html
- body does not contain json
- even if body was interpretable as json, it does not contain issuer, jwks_uri, authorization_endpoint and all other typical keys in the response

I expected the health check to fail since the Identity Server URL is not properly answered by an Identity Server.

```
                healthCheckBuilder.AddIdentityServer(identityServerUri);
```

Use an identityServerUri that responds 200 for all paths under it, but serves e.g. HTML instead of JSON. The path ".well-known/openid-configuration" will be requested and the response will erroneously be accepted by the Identity Server health checker.

**Environment**:

- .NET Core version: 3.1
- Healthchecks version: 3.1.1 (but also as of [current master](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/ca3b043/src/HealthChecks.IdSvr/IdSvrHealthCheck.cs))

cieciurm:
- This PR adds validation of OpenID Connect Discover response
- Validation follows rules from [the spec](https://openid.net/specs/openid-connect-discovery-1_0.html#IssuerDiscovery)

closes #701

- This is a very early PR
- I want to ask for feedback on the approach
- Tests will be added when we agree on solution

- No

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing


cieciurm:
Thanks for any suggestions and feedback 😄 
cc @sungam3r

src/HealthChecks.OpenIdConnectServer/OidcValidationHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    private static bool AnyValueContains(IEnumerable<string> values, IEnumerable<string> requiredValues) =>
+        values.Any(v => requiredValues.Contains(v));
+
+    private static string GetMissingValueExceptionMessage(string value) =>
+        $"Invalid discover response - '{value}' must be set!";
+
+    private static string GetMissingRequiredValuesExceptionMessage(string value, IEnumerable<string> requiredValues) =>
+        $"Invalid discover response - '{value}' must contain {string.Join(",", requiredValues)}!";
```
sungam3r, line 28:
```suggestion
        $"Invalid discover response - '{value}' must be one of the following values: {string.Join(",", requiredValues)}!";
```

src/HealthChecks.OpenIdConnectServer/OidcConstants.cs
```diff
@@ -0,0 +1,22 @@
...
+
+    internal const string SUBJECT_TYPES_SUPPORTED = "subject_types_supported";
+
+    internal const string ALGORITHMS_SUPPORTED = "id_token_signing_alg_values_supported";
+
+    internal static IEnumerable<string> REQUIRED_RESPONSE_TYPES => new[] { "code", "id_token", "token id_token" };
+
+    internal static IEnumerable<string> REQUIRED_SUBJECT_TYPES => new[] { "pairwise", "public" };
+
+    internal static IEnumerable<string> REQUIRED_ALGORITHMS => new[] { "RS256" };
```
sungam3r, line 21:
```suggestion
    internal static string[] REQUIRED_RESPONSE_TYPES => new[] { "code", "id_token", "token id_token" };

    internal static string[] REQUIRED_SUBJECT_TYPES => new[] { "pairwise", "public" };

    internal static string[] REQUIRED_ALGORITHMS => new[] { "RS256" };
```

src/HealthChecks.OpenIdConnectServer/IdSvrHealthCheck.cs
```diff
@@ -27,13 +28,35 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
...
+
+            return HealthCheckResult.Healthy();
         }
         catch (Exception ex)
         {
             return new HealthCheckResult(context.Registration.FailureStatus, exception: ex);
         }
     }
+
+    private static void ValidateResponse(DiscoveryEndpointResponse response)
```
sungam3r, line 52: Let's make it an instance method of `DiscoveryEndpointResponse` + place https://openid.net/specs/openid-connect-discovery-1_0.html#IssuerDiscovery link in comments for that method.

src/HealthChecks.OpenIdConnectServer/DiscoveryEndpointResponse.cs
```diff
@@ -0,0 +1,24 @@
...
+    public string JwksUri { get; set; } = null!;
+
+    [JsonPropertyName(OidcConstants.RESPONSE_TYPES_SUPPORTED)]
+    public IEnumerable<string> ResponseTypesSupported { get; set; } = null!;
+
+    [JsonPropertyName(OidcConstants.SUBJECT_TYPES_SUPPORTED)]
+    public IEnumerable<string> SubjectTypesSupported { get; set; } = null!;
+
+    [JsonPropertyName(OidcConstants.ALGORITHMS_SUPPORTED)]
+    public IEnumerable<string> IdTokenSigningAlgValuesSupported { get; set; } = null!;
```
sungam3r, line 23:
```suggestion
    public string[] ResponseTypesSupported { get; set; } = null!;

    [JsonPropertyName(OidcConstants.SUBJECT_TYPES_SUPPORTED)]
    public string[] SubjectTypesSupported { get; set; } = null!;

    [JsonPropertyName(OidcConstants.ALGORITHMS_SUPPORTED)]
    public string[] IdTokenSigningAlgValuesSupported { get; set; } = null!;
```

sungam3r:
> I want to ask for feedback on the approach
Tests will be added when we agree on solution

Looks good thought I have not read the spec.

src/HealthChecks.OpenIdConnectServer/DiscoveryEndpointResponse.cs
cieciurm: Could you share more thoughts why array is better than IEnumerable here?

src/HealthChecks.OpenIdConnectServer/OidcConstants.cs
cieciurm: Same question as above. Here I am using only `Contains(...)` method  so I thought I would use the most most basic collection interface.

src/HealthChecks.OpenIdConnectServer/DiscoveryEndpointResponse.cs
sungam3r: 1. Nothing to abstract here. Internal code.
2. Abstraction gives nothing except additional iterator allocation when using `Contains` etc.

src/HealthChecks.OpenIdConnectServer/OidcConstants.cs
sungam3r: >  so I thought I would use the most most basic collection interface.

Indeed, but it's not needed when working in internal (i.e. 100% your) code.

src/HealthChecks.OpenIdConnectServer/DiscoveryEndpointResponse.cs
cieciurm: Understood, thanks!

src/HealthChecks.OpenIdConnectServer/IdSvrHealthCheck.cs
cieciurm: Thanks, done.

src/HealthChecks.OpenIdConnectServer/DiscoveryEndpointResponse.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    private static void ValidateValue(string value, string metadata)
+    {
+        if (string.IsNullOrWhiteSpace(value))
+        {
+            throw new ArgumentException(GetMissingValueExceptionMessage(metadata));
+        }
+    }
+
+    private static void ValidateRequiredValues(IEnumerable<string> values, string metadata, IEnumerable<string> requiredValues)
```
sungam3r, line 47:
```suggestion
    private static void ValidateRequiredValues(IEnumerable<string> values, string metadata, string[] requiredValues)
```

src/HealthChecks.OpenIdConnectServer/DiscoveryEndpointResponse.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        if (values == null || !AnyValueContains(values, requiredValues))
+        {
+            throw new ArgumentException(GetMissingRequiredValuesExceptionMessage(metadata, requiredValues));
+        }
+    }
+
+    private static bool AnyValueContains(IEnumerable<string> values, IEnumerable<string> requiredValues) =>
```
sungam3r, line 55:
```suggestion
    private static bool AnyValueContains(IEnumerable<string> values, string[] requiredValues) =>
```

src/HealthChecks.OpenIdConnectServer/DiscoveryEndpointResponse.cs
```diff
@@ -0,0 +1,63 @@
...
+        values.Any(v => requiredValues.Contains(v));
+
+    private static string GetMissingValueExceptionMessage(string value) =>
+        $"Invalid discovery response - '{value}' must be set!";
+
+    private static string GetMissingRequiredValuesExceptionMessage(string value, IEnumerable<string> requiredValues) =>
```
sungam3r, line 61:
```suggestion
    private static string GetMissingRequiredValuesExceptionMessage(string value, string[] requiredValues) =>
```

src/HealthChecks.OpenIdConnectServer/DiscoveryEndpointResponse.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    private static void ValidateRequiredValues(IEnumerable<string> values, string metadata, string[] requiredValues)
+    {
+        if (values == null || !AnyValueContains(values, requiredValues))
+        {
+            throw new ArgumentException(GetMissingRequiredValuesExceptionMessage(metadata, requiredValues));
+        }
+    }
+
+    private static bool AnyValueContains(IEnumerable<string> values, string[] requiredValues) =>
```
sungam3r, line 55:
```suggestion
    private static bool AnyValueContains(string[] values, string[] requiredValues) =>
```

sungam3r (approved):

src/HealthChecks.OpenIdConnectServer/DiscoveryEndpointResponse.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    private static void ValidateValue(string value, string metadata)
+    {
+        if (string.IsNullOrWhiteSpace(value))
+        {
+            throw new ArgumentException(GetMissingValueExceptionMessage(metadata));
+        }
+    }
+
+    private static void ValidateRequiredValues(IEnumerable<string> values, string metadata, string[] requiredValues)
```
sungam3r, line 47:
```suggestion
    private static void ValidateRequiredValues(string[] values, string metadata, string[] requiredValues)

```

cieciurm:
I would like to add tests for validation.

sungam3r:
Reminder - bump version.

cieciurm:
I've added applied code suggestions, added unit tests and bumped the version.

I decided to write more sophisticated check for `response_types_supported`. The spec mentiones a combined response type `token id_token`, but some identity providers (f.e. Identity Server and Azure AD) return `id_token token`. So I decided to validate separately the single response types (`code`, `id_token`) and separately combined response type. This should prevent false negatives.

cc @sungam3r

src/HealthChecks.OpenIdConnectServer/DiscoveryEndpointResponse.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// </summary>
+    public void ValidateResponse()
+    {
+        ValidateValue(Issuer, OidcConstants.ISSUER);
+        ValidateValue(AuthorizationEndpoint, OidcConstants.AUTHORIZATION_ENDPOINT);
+        ValidateValue(JwksUri, OidcConstants.JWKS_URI);
+
+        ValidateRequiredValues(ResponseTypesSupported, OidcConstants.RESPONSE_TYPES_SUPPORTED, OidcConstants.REQUIRED_RESPONSE_TYPES);
+
+        // Specification decribes 'token id_token' response type,
```
sungam3r, line 36:
```suggestion
        // Specification describes 'token id_token' response type,
```

src/HealthChecks.OpenIdConnectServer/HealthChecks.OpenIdConnectServer.csproj
```diff
@@ -11,6 +11,11 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Http" Version="7.0.0" />
     <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" Version="7.0.9" />
+    <PackageReference Include="System.Net.Http.Json" Version="7.0.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <InternalsVisibleTo Include="HealthChecks.OpenIdConnectServer.Tests" />
```
sungam3r, line 18: See #1987

cieciurm:
Thanks, `<InternalsVisibleTo />` has been removed. 

cc @sungam3r

sungam3r (approved):

PR #1996, opened 2023-08-14
Order health checks entries

valentinmarinro:
Order health checks entries by name to have a consistent list when adding new health checks in microservices 

When adding a new health check in any new or existing microservices, would be good to have them order by name

- [x ] Code compiles correctly


sungam3r:
I bumped version to 7.0.3 in 0bb15e9

PR #2028, opened 2023-09-05
Azure Key Vault Secrets health check

adamsitnik:
Hello!

My name is Adam, I work on the .NET Team where we are currently working on a set of libraries for cloud native applications. I cannot say much about the project, besides the fact that it’s going to use health checks.

To foster the .NET OSS Ecosystem, we would like to take a dependency on the `AspNetCore.Diagnostics.HealthChecks` project and provide PRs for things that we need.

One of them is an Azure Key Vault Secrets health check that has no dependency on `Azure.Security.KeyVault.Certificates` and `Azure.Security.KeyVault.Keys` packages (to reduce the size when they are not used) and does not require a pre-existing secret like the existing Azure Key Vault health check does (so it's very easy to use and get started with):

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/c000704e1e256c932539fcd5d4c8df781e8ec1a7/src/HealthChecks.AzureKeyVault/AzureKeyVaultHealthCheck.cs#L33-L34

This PR provides such health check, tests and documentation.

Please excuse me for not creating the issue up-front, but I find it easier to just discuss something when code is provided.

cc @sungam3r @eerhardt @DamianEdwards

### DIFF
--- a/src/HealthChecks.Azure.KeyVault.Secrets/DependencyInjection/AzureKeyVaultHealthChecksBuilderExtensions.cs
+++ b/src/HealthChecks.Azure.KeyVault.Secrets/DependencyInjection/AzureKeyVaultHealthChecksBuilderExtensions.cs
@@ -1,0 +1,50 @@
+using Azure.Security.KeyVault.Secrets;
+using HealthChecks.Azure.KeyVault.Secrets;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+/// <summary>
+/// Extension methods to configure <see cref="AzureKeyVaultSecretsHealthCheck"/>.
+/// </summary>
+public static class AzureKeyVaultHealthChecksBuilderExtensions
+{
+    private const string HEALTH_CHECK_NAME = "azure_key_vault_secret";
+
+    /// <summary>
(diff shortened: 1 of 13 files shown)


unaizorrilla:
LGTM!

Thanks for contribute this! I will add the new pipeline for this project asap! We have a pipeline for each package for ci/cd/preview versions!

sungam3r:
What about naming convention? Single word vs multiple words
изображение

src/HealthChecks.Azure.KeyVault.Secrets/DependencyInjection/AzureKeyVaultHealthChecksBuilderExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+/// <summary>
+/// Extension methods to configure <see cref="AzureKeyVaultSecretsHealthCheck"/>.
+/// </summary>
+public static class AzureKeyVaultHealthChecksBuilderExtensions
+{
+    private const string HEALTH_CHECK_NAME = "azure_key_vault_secret";
```
sungam3r, line 12: All other health checks across repo do not use snake_case convention for healthcheck names.

PR #2035, opened 2023-09-14
fix RCS1090: add missing ConfigureAwait(false)

adamsitnik:
VS is producing a lot of warnings for `RCS1090`: the missing `ConfigureAwait(false)`

I've just hit alt+enter and applied it to the whole solution


unaizorrilla:
Just to know!  This package are just for ASP.NET Core. In this scenario ( with this scheduler ) configureAwait is really necessary?

adamsitnik:
> This package are just for ASP.NET Core. In this scenario ( with this scheduler ) configureAwait is really necessary?

@unaizorrilla You are right: https://devblogs.microsoft.com/dotnet/configureawait-faq/#when-should-i-use-configureawaitfalse

I've reverted the change for .exe(s) and kept it for libs.

Once the CI gets green I am going to squash the PR to get to 0 warnings in #2044

sungam3r:
I didn't force `ConfigureAwait(false)` for UI-projects intentionally for a long time but I'm fine if someone eventually did.

PR #2036, opened 2023-09-14
try to introduce .yml templates so introducing new packages requires less work

adamsitnik:
@unaizorrilla @lfraile I am about to introduce more new packages and I wonder whether it would be possible to reduce the .yml duplication by using templates.

This is just a DRAFT, I've not used .yml in a long time.


unaizorrilla:
Hi @lfraile 

Please, can you take a look on this one?

adamsitnik:
Damn, I've used AzDO syntax, I'll rewrite it to GH Actions and ping you once the CI becomes green.

PR #2037, opened 2023-09-14
use GH reusable workflows to reduce .yml duplication

adamsitnik:
I am about to introduce more new packages and I wonder whether it would be possible to reduce the .yml duplication by using workflows.

I wrote it based on https://dev.to/n3wt0n/avoid-duplication-github-actions-reusable-workflows-3ae8

### DIFF
--- a/.github/workflows/healthchecks_azurekeyvault_cd.yml
+++ b/.github/workflows/healthchecks_azurekeyvault_cd.yml
@@ -8,22 +8,8 @@ on:
 
 jobs:
   build:
-    env:
+    uses: ./.github/workflows/reusable_workflow.yml
+    with:
       BUILD_CONFIG: Release
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - name: Setup .NET
-        uses: actions/setup-dotnet@v3
-        with:
-          dotnet-version: |
-            6.0.x
-            7.0.x
-      - name: Restore
-        run: dotnet restore ./src/HealthChecks.AzureKeyVault/HealthChecks.AzureKeyVault.csproj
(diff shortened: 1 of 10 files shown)


adamsitnik:
@lfraile it's ready for review, PTAL

.github/workflows/healthchecks_azurekeyvault_cd_preview.yml
```diff
@@ -8,23 +8,9 @@ on:
 
 jobs:
   build:
-    env:
+    uses: ./.github/workflows/reusable_cd_preview_workflow.yml
```
lfraile, line 11:
```suggestion
    uses: ./.github/workflows/reusable_cd_preview_workflow.yml
    secrets: inherit
```
We will need this so we can use ${{secrets.NUGET_API_KEY}}

.github/workflows/healthchecks_azurekeyvault_cd.yml
```diff
@@ -8,22 +8,8 @@ on:
 
 jobs:
   build:
-    env:
+    uses: ./.github/workflows/reusable_cd_workflow.yml
```
lfraile, line 11:
```suggestion
    uses: ./.github/workflows/reusable_cd_workflow.yml
    secrets: inherit
```
We will need this so we can use ${{secrets.NUGET_API_KEY}}

lfraile (changes requested):

adamsitnik:
@lfraile great catch! I've applied your suggestion to all files, PTAL

lfraile (approved):

sungam3r:
@adamsitnik Thank you very much for that!

PR #2038, opened 2023-09-14
split Azure packages into dedicated nuget packages

### LINKED ISSUE #665: Azure Service Bus Queue and Azure Queue Storage registrations use the same name
Please, fill the following sections to help us fix the issue

**What happened**:

These two health checks use the same name for the corresponding registrations

To use different names

[AddAzureQueueStorage](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/96465d0424b8877378a7a3d74c1477e02bb7deed/src/HealthChecks.AzureStorage/DependencyInjection/AzureStorageHealthCheckBuilderExtensions.cs#L81)

[AddAzureServiceBusQueue](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/1acc622d698c80adb5272a92ee226176d7ee49a7/src/HealthChecks.AzureServiceBus/DependencyInjection/AzureServiceBusHealthCheckBuilderExtensions.cs#L58)

### LINKED ISSUE #1371: Leverage Existing Azure Service Clients Using Dependency Injection

### LINKED ISSUE #1567: HealthChecks.CosmosDB - Deferred CosmosClient resolution
**What would you like to be added:**

I'd like to a factory method that allows for resolution of the `CosmosClient` to be controlled similar to how the configureOptions is able to be resolved with the `IServiceProvider`. 

```csharp
        /// <summary>
        /// Add a health check for Azure CosmosDb database.
        /// </summary>
        /// <param name="builder">The <see cref="IHealthChecksBuilder"/>.</param>
        /// <param name="clientFactory">Delegate for creating the <see cref="CosmosClient"/> instance that will communicate with the database.</param>
        /// <param name="configureOptions">Delegate for configuring the health check. Optional.</param>
        /// <param name="name">The health check name. Optional. If <see langword="null"/> the type name 'cosmosdb' will be used for the name.</param>
        /// <param name="failureStatus">
        /// The <see cref="HealthStatus"/> that should be reported when the health check fails. Optional. If <see langword="null"/> then
        /// the default status of <see cref="HealthStatus.Unhealthy"/> will be reported.
        /// </param>
        /// <param name="tags">A list of tags that can be used to filter sets of health checks. Optional.</param>
        /// <param name="timeout">An optional <see cref="TimeSpan"/> representing the timeout of the check.</param>
        /// <returns>The specified <paramref name="builder"/>.</returns>
        public static IHealthChecksBuilder AddCosmosDb(
            Func<IServiceProvider, CosmosClient> clientFactory,
            Action<IServiceProvider, CosmosDbHealthCheckOptions>? configureOptions = default,
            string? name = default,
            TimeSpan? timeout = default)
        {
               name ?? COSMOS_NAME,
               sp =>
               {
                   var options = new CosmosDbHealthCheckOptions();
                   configureOptions?.Invoke(sp, options);
                   return new CosmosDbHealthCheck(clientFactory(sp), options);
               },
               failureStatus,
               tags,
               timeout));
        }

```

**Why is this needed:**

Our CosmosClients aren't always initialized as soon as the application starts up and we will have more than one because we have multiple cosmos DBs (so 1 client per DB, not per container).  We have a factory that lazy initializes them the first time they are needed.  Additionally the configuration for these clients is injected via DI initially.  As such we need to be able to defer resolution of the client.  By adding the ability to return the `CosmosClient` ourselves during runtime, we can ensure that the client we want is returned.

I'd be happy to go through the steps to get this in if it makes sense.

PS - Sorry for the 2nd issue on this.  I had made one for the documentdb package, but realized that was the legacy SDK.  I had hoped that this new package would work as it had deferred resolution of some things, but after digging through the source, it turned out it wasn't quite able to do what I needed.

### LINKED ISSUE #2040: Azure health checks: How to steer the users towards best practices
Azure SDK Clients are recommended to be used as singletons.

From https://devblogs.microsoft.com/azure-sdk/lifetime-management-and-thread-safety-guarantees-of-azure-sdk-net-clients/:

> The main rule of Azure SDK client lifetime management is: treat clients as singletons.

Most of the existing Azure Health Checks allow the user to specify a client instance. Example:

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/7060bdd2105f5cf53539a467504e9ed53d4d7e04/src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs#L24

But they also provide constructors that take the arguments required to create such client instance. Example:

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/7060bdd2105f5cf53539a467504e9ed53d4d7e04/src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs#L12

In such case, they are adding them to a static cache:

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/7060bdd2105f5cf53539a467504e9ed53d4d7e04/src/HealthChecks.AzureStorage/AzureBlobStorageHealthCheck.cs#L14

I believe that it leads to having at least two instances of the client created:
- one used by the app outside of the health chekc
- one used by the health check.

This is bad for perf, but also in some rare edge cases can potentially lead into a situation when the client used by the app is disconnected, but the client used by the HC is connected and HC returns "Healthy".

Moreover, caching is hard and having a custom cache can always lead into bugs that are hard to diagnose. And increased complexity that needs to be maintained.

I would like to propose following changes for health checks that use clients that should be used as singletons:
- each HC should expose only one constructor that takes an existing instance of the client (this forces the user to create them, takes complexity away from us).

Sample:

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/7060bdd2105f5cf53539a467504e9ed53d4d7e04/src/HealthChecks.Azure.KeyVault.Secrets/AzureKeyVaultSecretsHealthCheck.cs#L31

- removal of the static cache (less complexity, better size on disk, smaller chance for bugs)
- each HC should expose only one DI extension method. The method should by default resolve an instance of the client from `IServiceProvider`, but let the user override this behaviour. 

Sample:

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/7060bdd2105f5cf53539a467504e9ed53d4d7e04/src/HealthChecks.Azure.KeyVault.Secrets/DependencyInjection/AzureKeyVaultHealthChecksBuilderExtensions.cs#L32-L49

This would let us steer the users towards best practices:
- the users are responsible for registering the client in the DI (using whatever ctor they choose)
- HC reuse the existing singleton instance

This should improve perf (only one instance per app) and complexity (very simple).
The downside is the introduction of breaking changes.

```csharp
public void Configure(IHealthChecksBuilder builder)
{
    builder.Services.AddSingleton(sp => new SecretClient(new Uri("azure-key-vault-uri"), new DefaultAzureCredential()));
    builder
        .AddHealthChecks()
        .AddAzureKeyVaultSecrets(); // by default resolves SecretClient from the service provider
}
```

For those who need more than one instance per app, we would update to .NET 8 and let them user the keyed DI:

```csharp
public void Configure(IHealthChecksBuilder builder)
{
    builder.Services
        .AddKeyedSingleton(serviceKey: "first", (serviceProvider, serviceKey) => new SecretClient(new Uri("azure-key-vault-uri-first"), new DefaultAzureCredential()))
        .AddKeyedSingleton(serviceKey: "second", (serviceProvider, serviceKey) => new SecretClient(new Uri("azure-key-vault-uri-second"), new DefaultAzureCredential()))
        .AddHealthChecks()
            .AddAzureKeyVaultSecrets(secretClientFactory: serviceProvider => serviceProvider.GetRequiredKeyedService<SecretClient>("first"), healthCheckName: "secrets_first")
            .AddAzureKeyVaultSecrets(secretClientFactory: serviceProvider => serviceProvider.GetRequiredKeyedService<SecretClient>("second"), healthCheckName: "secrets_second");
}
```

@unaizorrilla @sungam3r what is your opinion on that?

adamsitnik:
This PR contains following changes::

- [x] split HealthChecks.AzureStorage into Storage, Files and Queues packages
- [x] remove HealthChecks.AzureStorage 
- [x] move Azure Tables health check from HealthChecks.CosmosDb to dedicated package
- [x] move Azure Event Hubs health check from HealthChecks.AzureServiceBus to dedicated package
- [x] enforce #2040
- [x] add missing README.md

What is not included in this PR (I'll send a separate PR after this gets merged):
- sign new assemblies
- update to 8.0 microsoft extensions

fixes #2040
fixes #1371 (health check reuse registered clients rather than creating new onesS)
fixes #1567 (client instance is created lazily when needed for the first time)
fixes #665 (two health checks were using the same name)


unaizorrilla:
Hi @adamsitnik 

The idea is mantain the AzureStorage package with 3 references to Tables, Queues and Shares? Why not deprecated current Storage and move the IHealthCheckBuilder extensiosn to each package?

build/versions.props
```diff
@@ -28,7 +33,7 @@
     <HealthCheckInfluxDB>7.0.0</HealthCheckInfluxDB>
     <HealthCheckIoTHub>7.0.0</HealthCheckIoTHub>
     <HealthCheckKafka>7.0.0</HealthCheckKafka>
-    <HealthCheckKeyVault>7.0.0</HealthCheckKeyVault>
+    <HealthCheckKeyVault>7.0.1</HealthCheckKeyVault>
```
adamsitnik, line 36:
```suggestion
    <HealthCheckKeyVault>7.0.0</HealthCheckKeyVault>
```

src/HealthChecks.Azure.Messaging.EventHubs/README.md
```diff
@@ -0,0 +1,42 @@
...
+    builder.Services
+        .AddKeyedSingleton(serviceKey: "eventHubName1", (serviceProvider, serviceKey) => new EventHubProducerClient("fullyQualifiedNamespace", "eventHubName1", new DefaultAzureCredential()))
+        .AddKeyedSingleton(serviceKey: "eventHubName2", (serviceProvider, serviceKey) => new EventHubProducerClient("fullyQualifiedNamespace", "eventHubName2", new DefaultAzureCredential()))
+        .AddHealthChecks()
+            .AddAzureKeyVaultSecrets(clientFactory: serviceProvider => serviceProvider.GetRequiredKeyedService<EventHubProducerClient>("eventHubName1"), healthCheckName: "event_hub_1")
+            .AddAzureKeyVaultSecrets(clientFactory: serviceProvider => serviceProvider.GetRequiredKeyedService<EventHubProducerClient>("eventHubName2"), healthCheckName: "event_hub_2");
+
+}
+```
+
```
adamsitnik, line 38:
```suggestion
```

Issue #1993, opened 2023-08-11
Too many open connections with NpgSql 7.0.0


peterkiss1:
**What happened**: upgraded from 6.0.2 to 7.0.0 and 1 application instance opens ~40 connections towards the database which cause issues as the number of connections are limited, with the old version it has opened ~2 connections.

**What you expected to happen**: should not have this many connection opened and left dangling.

**How to reproduce it**: grab an application (checked with .NET 7 and 8 Preview) add this package and start running the health checks.

**Source code sample**: no special code is required, plain vanilla ASP.NET health check setup.

**Environment**:

- .NET Core version: 7 and 8 Preview
- Healthchecks version: NpgSql 7.0.0
- Operative system: Linux

tyshkavets:
Experiencing the same thing. Seems to be due to an introduction of DbDataSource abstraction in .NET 7 [(details here)](https://github.com/dotnet/runtime/issues/64812) and `NpgSqlDataSource` implementation for it in corresponding NpgSql version.

According to `NpgSqlHealthCheckBuilderExtensions.cs`, a new one is spawned every time we execute a health check, leading to excessive numbers of dangling connections. It happens because the code responsible for creating DbDataSource is simply packed into a delegate and passed into `HealthCheckRegistration.Factory`, to be invoked on every `RunCheckAsync()`. 

It's a rather unpleasant surprise, I'd (perhaps, naively) expect them to be pooled under the hood.

My workaround was to create a data source, stick it as a singleton into DI and use `.AddNpgSql()` overload that accepts `Func<IServiceProvider, NpgsqlDataSource>` rather than a connection string, effectively pooling it on my own. Should be fine if your connection string isn't prone to changing.

sungam3r:
OK, got it. Need time to think how to deal with it.

iam-black:
I encountered the same issue and used the same overload as @tyshkavets for the fix, but without explicit singleton registration:
`var options = new NpgSqlHealthCheckOptions
{
    DataSource = new NpgsqlDataSourceBuilder(connectionString).Build(),
};
hcBuilder.AddNpgSql(options);`

kamisoft-fr:
Hi, 
I had the same issue, and I finally ended up by tweaking the connection string to avoid pooling:

```
            string connectionString = configuration.GetConnectionString("xxx") ;

            NpgsqlConnectionStringBuilder healthCheckConnectionStringBuilder = new(connectionString)
[...]

PR #2045, opened 2023-09-21
Npgsql: don't create more than one DataSource

adamsitnik:
3 out of 4 DI extension methods are now ensuring that the `DataSource` is created only once (in a thread-safe way), and when it's possible they reuse the instance registered in the DI.

It's not an ideal fix, because the current design still allows the users to have more than one `NpgsqlDataSource` per their app. Example:

```cs
void Configure(IHostApplicationBuilder builder)
{
    builder.Services.AddNpgsqlDataSource("connectionString"); // registers a singleton factory for NpgsqlDataSource using Npgsql.DependencyInjection package, is used by the app
    builder.AddHealthChecks().AddNpgSql(new NpgSqlHealthCheckOptions()
    {
        ConnectionString = "connectionString" // registers a factory used only by the health check
    });
}
```

@unaizorrilla it's the fix I propose for 7.0.

For 8.0 we need to introduce a breaking change similar to what I did for the Azure clients in https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/2040

Contributes to #1993.

### DIFF
--- a/src/HealthChecks.NpgSql/HealthChecks.NpgSql.csproj
+++ b/src/HealthChecks.NpgSql/HealthChecks.NpgSql.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net7.0</TargetFrameworks>
     <PackageTags>$(PackageTags);Beat;Postgress</PackageTags>
     <Description>HealthChecks.NpgSql is a health check for Postgress Sql.</Description>
     <VersionPrefix>$(HealthCheckNpgSql)</VersionPrefix>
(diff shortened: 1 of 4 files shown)


src/HealthChecks.NpgSql/HealthChecks.NpgSql.csproj
```diff
@@ -1,14 +1,14 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net7.0</TargetFrameworks>
```
sungam3r, line 4: Why net7?

adamsitnik:
> Current fix looks like more a hack than general-purpose solution.

I agree, that is why I wrote:

> For 8.0 we need to introduce a breaking change similar to what I did for the Azure clients in

Related: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2096

src/HealthChecks.NpgSql/HealthChecks.NpgSql.csproj
sungam3r: Ok

PR #2048, opened 2023-09-26
Fix connections leak in RabbitMQHealthCheck

Vladimirezh:
Every call of healthcheck create new connection. Connection doesn't close. 

No

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.Rabbitmq/DependencyInjection/RabbitMQHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.Rabbitmq/DependencyInjection/RabbitMQHealthCheckBuilderExtensions.cs
@@ -152,10 +152,10 @@ public static class RabbitMQHealthCheckBuilderExtensions
     {
         var options = new RabbitMQHealthCheckOptions();
         setup?.Invoke(options);
-
+        builder.Services.AddSingleton(new RabbitMQHealthCheck(options));
         return builder.Add(new HealthCheckRegistration(
             name ?? NAME,
-            new RabbitMQHealthCheck(options),
+            sp => sp.GetRequiredService<RabbitMQHealthCheck>(),
             failureStatus,
             tags,
             timeout));
@@ -187,15 +187,15 @@ public static class RabbitMQHealthCheckBuilderExtensions
         TimeSpan? timeout = default)
     {
         var options = new RabbitMQHealthCheckOptions();
+        builder.Services.AddSingleton(sp =>
+        {
+            setup?.Invoke(sp, options);
+            return new RabbitMQHealthCheck(options);
+        });
 
         return builder.Add(new HealthCheckRegistration(
             name ?? NAME,
-            sp =>
-            {
-                setup?.Invoke(sp, options);
-
-                return new RabbitMQHealthCheck(options);
-            },
+            sp => sp.GetRequiredService<RabbitMQHealthCheck>(),
             failureStatus,
             tags,
             timeout));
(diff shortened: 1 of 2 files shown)


Vladimirezh:
@dotnet-policy-service agree

unaizorrilla:
Hi @Vladimirezh 

Let me some days to review.

unaizorrilla:
Hi @Vladimirezh 

Thanks for submit this PR, just a few notes

- Just tested the current code on sample app HealthChecks.Sample with Rabbit and just one connection is created and the health check is reused always using the uri and options overloads
- You can register your connection and connection factory on the container and use the proper extension method to alternate registration method

Vladimirezh:
To reproduce leak modify ConfigureServices of HealthChecks.Sample to:
```
  //Custom rabbitmq factory
        services.AddSingleton<Func<IConnection>>(() =>
        {
            var factory = new ConnectionFactory
            {
                Uri = new Uri("amqp://localhost:5672"),
                AutomaticRecoveryEnabled = true,
                UseBackgroundThreadsForIO = true
            };

            return factory.CreateConnection();
        });


        services
            .AddApplicationInsightsTelemetry()
            .AddHealthChecks()
            // .AddRabbitMQ(rabbitConnectionString: "amqp://localhost:5672", name: "rabbit1"); //<- Work fine
            .AddRabbitMQ((sp, options) =>
                options.Connection = sp.GetRequiredService<Func<IConnection>>()()); //<- Leak
```
After call route /healthz many times look to Rabbit Management UI
image

unaizorrilla:
Hi @Vladimirezh 

Just curious about your code.. some notes...

If you registered the IConnection as a Singleton and use the overload you mention like this

```csharp
       
        var connectionFactory = new ConnectionFactory
        {
            Uri = new Uri("amqp://localhost:5672")
        };

        services.AddSingleton<IConnection>(connectionFactory.CreateConnection());

        services
            .AddHealthChecks()
            .AddRabbitMQ((sp, options) =>
                options.Connection = sp.GetRequiredService<IConnection>(), name: "rabbit1");
```

just one connection is used 

image


Why you need to register a Func of IConnection?

Vladimirezh:
I have custom connection factory. Factory doesn't implement IConnectionFactory. In example it's Func, in my project it's library abstraction. I use overload with setup and set connection to options.

sungam3r:
@Vladimirezh Proposed changes will not work for multiple registrations. You register RabbitMQHealthCheck as singleton in method accepting user defined `Action<RabbitMQHealthCheckOptions>? setup`/`Action<IServiceProvider, RabbitMQHealthCheckOptions>? setup`, `name`, etc. Second and all other registrations will fetch the latest registered `RabbitMQHealthCheck` from ServiceProvider. This is a well-known design flaw and codebase tends to avoid such usages although you still can find such code in some health checks (alas).

If you look in the code in `master` then you'll see that `RabbitMQHealthCheck` caches connections in static dictionary by options instance as as key. I think that the real problem here - calling `Action<IServiceProvider, RabbitMQHealthCheckOptions>? setup` inside `HealthCheckRegistration`. Other overload with `Action<RabbitMQHealthCheckOptions>? setup` calls it outside - one time. Why so? Because ServiceProvider is available only inside `HealthCheckRegistration` delegate so user-defined delegate is called each time. I agree that options should be changed (initialized) only once. In our overload it's initialized deferred and not once. Please check #2153. It should fix that issue. If OK I would like you to make subsequent PR with all your tests.

ping @adamsitnik It seems to me that we have similar problems with some other checks. Did you fix this with Npgsql or was there something else? I still have not returned to the full support of the project, so I have not studied your changes.

PR #2093, opened 2023-11-17
Send a MySQL ping packet by default

### LINKED ISSUE #2031: MySql check no longer uses Ping

Before https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/1835, `MySqlHealthCheck` used `PingAsync` to check the server health. It now defaults to `SELECT 1;`.

I would like to be able to opt in to using a ping, or even having that restored as the default behavior.

I propose setting the default `MySqlHealthCheckOptions.CommandText` value to `null`, and having `MySqlHealthCheck` interpret that as meaning "ping the server". If `CommandText` is non-`null`, then it would be used to form a `MySqlCommand`.

**Why is this needed**:

A `COM_PING` packet will be much more lightweight to send to the server instead of executing a query, even a simple one.

bgrainger:
The previous default behavior for the MySQL health check was to execute `SELECT 1;`. This is fairly heavyweight; it would be more efficient to send a MySQL "ping" packet, which is a binary test of server health. `MySqlHealthCheckOptions.CommandText` now defaults to `null`, meaning "ping the server". Users can opt in to setting a command that will be executed on the server, but the default is now a more efficient ping.

Fixes #2031

Please reference the issue this PR will close: #2031

Yes; it makes `MySqlHealthCheckOptions.CommandText` nullable and changes its default value; it also makes the `healthQuery` parameter to `AddMySql` nullable and changes its default value to `null`.

- [X] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.MySql/DependencyInjection/MySqlHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.MySql/DependencyInjection/MySqlHealthCheckBuilderExtensions.cs
@@ -10,14 +10,13 @@ namespace Microsoft.Extensions.DependencyInjection;
 public static class MySqlHealthCheckBuilderExtensions
 {
     private const string NAME = "mysql";
-    internal const string HEALTH_QUERY = "SELECT 1;";
 
     /// <summary>
     /// Add a health check for MySQL databases.
     /// </summary>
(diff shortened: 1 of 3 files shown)


adamsitnik:
The fix got included in #2096

PR #2096, opened 2023-11-18
Use MySqlDataSource in `AddMySql` by default

bgrainger:

This pushes users towards best practices (similar to what @adamsitnik described in #2040). The user should register a `MySqlDataSource` with DI, then call `.AddHealthChecks().AddMySql()` to use that same `MySqlDataSource` by default.

This PR also builds on top of #2093. They could be separated if needed, but there would be a merge conflict if they were opened separately, and both changes are useful.

Yes, the `.AddMySql` method now has an overload with all default parameters that "does the right thing".

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

I didn't see a place to update documentation or samples for MySql, but I'd be happy to do this to show how to add this health check properly.


src/HealthChecks.MySql/MySqlHealthCheckOptions.cs
```diff
@@ -10,14 +9,19 @@ namespace HealthChecks.MySql;
 public class MySqlHealthCheckOptions
 {
     /// <summary>
-    /// The MySQL connection string to be used.
+    /// The MySQL data source to be used. This is the preferred way to specify the MySQL server to be checked.
     /// </summary>
-    public string ConnectionString { get; set; } = null!;
+    public MySqlDataSource? DataSource { get; set; }
```
adamsitnik, line 14: Exposing a `DataSource` can lead to issues similar to the ones I've fixed in #2045 for PostgreSQL. IMO we should make it internal.

```suggestion
    internal MySqlDataSource? DataSource { get; set; }
```

test/HealthChecks.MySql.Tests/Functional/MySqlHealthCheckTests.cs
```diff
@@ -1,11 +1,39 @@
...
-    public async Task be_healthy_when_mysql_server_is_available()
+    public async Task be_healthy_when_mysql_server_is_available_using_data_source()
+    {
+        var connectionString = "server=localhost;port=3306;database=information_schema;uid=root;password=Password12!";
+
+        var webHostBuilder = new WebHostBuilder()
+            .ConfigureServices(services =>
+            {
+                services
+                    .AddMySqlDataSource(connectionString)
```
adamsitnik, line 17: Is there any chance you could add a `README.md` file similar to https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.Azure.Data.Tables/README.md that recommends using this health check with the registration of data source like this?

adamsitnik (commented):
@bgrainger thank you for your contribution! I believe it's the step in the right direction, it needs only some polishing before getting merged.

adamsitnik:
@bgrainger I've created https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/2113, PTAL and let me know what do you think

adamsitnik (commented):
It's almost ready, PTAL at my comments. Thank you!

src/HealthChecks.MySql/MySqlHealthCheckOptions.cs
bgrainger: > Let's make the parameterless MySqlHealthCheckOptions ctor internal ... Moreover, let's make the setters for these properties internal.

When developing this, I found that I didn't need a parameterless constructor nor to set the properties after construction. Thus, the `MySqlHealthCheckOptions` type has been updated to have just two public constructors and private setters for all properties.

adamsitnik: > When developing this, I found that I didn't need a parameterless constructor nor to set the properties after construction. Thus, the MySqlHealthCheckOptions type has been updated to have just two public constructors and private setters for all properties.

Great! 👍

src/HealthChecks.MySql/MySqlHealthCheck.cs
```diff
@@ -12,8 +12,11 @@ public class MySqlHealthCheck : IHealthCheck
...
     public MySqlHealthCheck(MySqlHealthCheckOptions options)
     {
-        Guard.ThrowIfNull(options.ConnectionString, true);
-        Guard.ThrowIfNull(options.CommandText, true);
+        Guard.ThrowIfNull(options);
+        if (options.DataSource is null && options.ConnectionString is null)
+            throw new InvalidOperationException("One of options.DataSource or options.ConnectionString must be specified.");
+        if (options.DataSource is not null && options.ConnectionString is not null)
+            throw new InvalidOperationException("Only one of options.DataSource or options.ConnectionString must be specified.");
         _options = options;
```
adamsitnik, line 20: We don't need these checks anymore, as the public ctors ensure that the state is always valid.

```suggestion
        _options = Guard.ThrowIfNull(options);
```

adamsitnik (approved):
LGTM, thank you very much @bgrainger !

src/HealthChecks.MySql/MySqlHealthCheck.cs
adamsitnik: For some reason I can't apply this suggestion. I am going to merge it now since it's not blocking

bgrainger: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2129

PR #2112, opened 2023-12-04
revert the removal of net5.0 moniker

### LINKED ISSUE #2103: Revert changes for net5 tfm
> for some projects I've removed net5.0 tfm as it's [not supported anymore](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core)

изображение

It should be reverted. It make no sence to require higher TFM in case no specific APIs are used from that TFM. Yes, net5 not supported anymore but this is story about runtime/end-applications but not about libraries. With such change as at picture above you make impossible for net5 apps to use this health check though healthcheck itself works perfectly targeted for net5.

_Originally posted by @sungam3r in https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/2044#issuecomment-1837560421_

adamsitnik:
fixes #2103

### DIFF
--- a/src/HealthChecks.Network/HealthChecks.Network.csproj
+++ b/src/HealthChecks.Network/HealthChecks.Network.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net6.0;net7.0;net8.0</TargetFrameworks>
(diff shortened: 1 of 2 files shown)


sungam3r:
Was it single project?

sungam3r (approved):

sungam3r (approved):
Is that only change you did when removing net5?

adamsitnik:
Great question! I've searched for usages of `#if` and found one more project that required the change (`HealthChecks.Network`)

Issue #2113, opened 2023-12-04
DB health checks: design


adamsitnik:
.NET 7 has introduced the concept of `DbDataSource`, from https://github.com/dotnet/runtime/issues/64812 we can read it's benefits:

> * DbDataSource encapsulates all information and configuration needed to provide an open database connection (auth information, logging setup, type mapping config, etc.), ready for executing commands.
> * This makes it suitable for passing around and registering in DI as a factory for connections, without needing to pass around any additional information.
> * It's intended for DbDataSources to correspond to connection pools managed internally in the database driver. 
>    * With the current API, drivers need to look up the internal pool each time a connection is opened, by the connection string. Since DbDataSource allows user code to directly reference an (abstracted) internal pool, it eliminates that lookup and helps perf.
>    * DbDataSource isn't an abstraction for a connection pool (see [#24856](https://github.com/dotnet/runtime/issues/24856)); that would require introducing some pooling-specific APIs, is complicated, and might not be a good idea. However, it could evolve into one in the future, with external pooling implementations which can be used across ADO.NET providers.
> * It's also possible to get and execute a DbCommand directly on DbDataSource, without needing to deal with DbConnection at all. In scenarios where no connection state is required (e.g. transactions), it's not necessary to burden users with managing the connection, so this can be done under the hood. In addition, this opens up command execution models which aren't tied to a specific connection (e.g. multiplexing on Npgsql).
> * The proposal includes a shim to make the new abstraction immediately available on all ADO.NET drivers, without them needing to implement anything (unless they wish to customize behavior).

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/1993 has proved, that our current design can lead into failures: when the users don't provide an existing `DbDataSource` when registering a health check, but just a `ConnectionString`, then an instance of `DbDataSource` is created each time the health check is invoked.
This is bad for performance, because we end up having multiple connections open. The fix that I've provided in #2045 was just a workaround, we need a proper solution that can be applied to all DB-based health checks (see #2096 as an example).

Similarly to https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/2040, we should steer the users towards best practices and provide an API that makes it very hard to fall into such perf traps.
We are soon about to release 8.0 and it's great opportunity to introduce the breaking change. Having README.md should help us to ease the pain.

Possible solutions:

[...]

PR #2116, opened 2023-12-04
Improve Npgsql health check design

adamsitnik:
This PR implements #2113 for Npgsql:

- offer two public `NpgSqlHealthCheckOptions` ctors: one that requires `ConnectionString` and another that requires `DataSource`. This allows us to ensure that every instance of this type is valid.
- make `ConnectionString` and `DataSource` properties readonly. This allows us to ensure that the customers are not providing both.
- add README.md and explain what we recommend and why

### DIFF
--- a/src/HealthChecks.NpgSql/README.md
+++ b/src/HealthChecks.NpgSql/README.md
@@ -1,0 +1,47 @@
+## PostgreSQL Health Check
+
+This health check verifies the ability to communicate with [PostgreSQL](https://www.postgresql.org/). It uses the [Npgsql](https://www.npgsql.org/) library.
+
+## NpgsqlDataSource
+
+Starting with Npgsql 7.0 (and .NET 7), the starting point for any database operation is [NpgsqlDataSource](https://www.npgsql.org/doc/api/Npgsql.NpgsqlDataSource.html). The data source represents your PostgreSQL database, and can hand out connections to it, or support direct execution of SQL against it. The data source encapsulates the various Npgsql configuration needed to connect to PostgreSQL, as well as the **connection pooling which makes Npgsql efficient**.
+
+Npgsql's **data source supports additional configuration beyond the connection string**, such as logging, advanced authentication options, type mapping management, and more.
+
+## Recommended approach
+
+To take advantage of the performance `NpgsqlDataSource` has to offer, it should be used as a **singleton**. Otherwise, the app might end up with having multiple data source instances, all of which would have their own connection pools. This can lead to resources exhaustion and major performance issues (Example: [#1993](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/1993)).
+
+We encourage you to use [Npgsql.DependencyInjection](https://www.nuget.org/packages/Npgsql.DependencyInjection) package for registering a singleton factory for `NpgsqlDataSource`.  It allows easy configuration of your Npgsql connections and registers the appropriate services in your DI container.
+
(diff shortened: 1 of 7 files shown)


src/HealthChecks.NpgSql/README.md
```diff
@@ -0,0 +1,47 @@
+## PostgreSQL Health Check
+
+This health check verifies the ability to communicate with [PostgreSQL](https://www.postgresql.org/). It uses the [Npgsql](https://www.npgsql.org/) library.
+
+## NpgsqlDataSource
```
adamsitnik, line 5: @roji @NinoFloris PTAL at what I wrote here and let me know if this is correct

adamsitnik:
I was unable to get a review, so I've eyeballed it multiple times. It LGTM, so I am merging it. In case I am wrong, we are going to release a fix later.

src/HealthChecks.NpgSql/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Starting with Npgsql 7.0 (and .NET 7), the starting point for any database operation is [NpgsqlDataSource](https://www.npgsql.org/doc/api/Npgsql.NpgsqlDataSource.html). The data source represents your PostgreSQL database, and can hand out connections to it, or support direct execution of SQL against it. The data source encapsulates the various Npgsql configuration needed to connect to PostgreSQL, as well as the **connection pooling which makes Npgsql efficient**.
```
roji, line 7: I'd maybe go softer and say "the recommended starting point", as connection strings are still fully supported etc.

src/HealthChecks.NpgSql/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+```csharp
+void Configure(IServiceCollection services)
+{
+    services.AddNpgsqlDataSource("Host=pg_server;Username=test;Password=test;Database=test");
+    services.AddHealthChecks().AddNpgSql();
+}
+```
+
+By default, the `NpgsqlDataSource` instance is resolved from service provider. If you need to access more than one PostgreSQL database, you can use keyed DI services to achieve that:
```
roji, line 27: :+1:

src/HealthChecks.NpgSql/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+We encourage you to use [Npgsql.DependencyInjection](https://www.nuget.org/packages/Npgsql.DependencyInjection) package for registering a singleton factory for `NpgsqlDataSource`.  It allows easy configuration of your Npgsql connections and registers the appropriate services in your DI container.
+
+To make the shift to `NpgsqlDataSource` as easy as possible,  the `Npgsql.DependencyInjection` package registers not just a factory for the data source, but also factory for `NpgsqlConnection`  (and even `DbConnection`). So, your app does not need to suddenly start using `NpgsqlDataSource` everywhere.
+
+```csharp
+void Configure(IServiceCollection services)
+{
+    services.AddNpgsqlDataSource("Host=pg_server;Username=test;Password=test;Database=test");
+    services.AddHealthChecks().AddNpgSql();
```
roji, line 23: Yep, that's the ideal way things should work - the data source is registered independently as a singleton in DI (via Npgsql.DependencyInjection), and then is implicitly picked up by anyone needing it (health checks, EF Core...). A bit similar to an ILoggerFactory.

roji (commented):
Sorry for the super late review, but overall everything here seems good! Yeah, the important thing is to have health checks automatically resolve a singleton NpgsqlDataSource from DI (previously registered via Npgsql.DependencyInjection), and for other cases, to provide ways to register with either a provided data source or a connection string.

PR #2126, opened 2023-12-15
Set preview version for mysql ui storage package

unaizorrilla:

Fix mysql ui storage package as preview because pomelo is on preview also


adamsitnik (approved):
LGTM, again thank you for a very quick fix @unaizorrilla !

Issue #2175, opened 2024-03-10
OpenIdConnectServer health check does not allow RS512 only signing algorithm


m-gug:
**What happened**:
The current implementation expects the OIDC provider to support RS256 as a signature algorithm.

**What you expected to happen**:
It should also be allowed that the OIDC provider only supports RS512 and the health check still works.

**How to reproduce it (as minimally and precisely as possible)**:
Run the health check against a ODIC Provider that only supports RS512 as signing algorithm.

- .NET Core version 8
- Healthchecks version 8.0.0

DanRogers99:
We also use another signing algorithm.

It would be good if we could override this property

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/73abc7a094f9ef625b83455898a21d22d2fa6a82/src/HealthChecks.OpenIdConnectServer/OidcConstants.cs#L23

m-gug:
@DanRogers99 
Yes, I agree. I added to my PR so that the algorithms can be specified as parameters.

PR #2176, opened 2024-03-10
Allow RS512 as valid signing algorithm

m-gug:
The current implementation of the OpenIdConnect health check requires that the server supports RS256 as a signing algorithm.
If the server only supports RS512, the health check will fail.
This PR extends the checking of permitted signature algorithms to include RS512 (others can be added to the constant).

#2175 

The test [be_healthy_if_idsvr_is_available](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/test/HealthChecks.OpenIdConnectServer.Tests/Functional/IdSvrHealthCheckTests.cs) fails in my setup - i guess it needs a running OIDC Server?

No

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.OpenIdConnectServer/DependencyInjection/IdSvrHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.OpenIdConnectServer/DependencyInjection/IdSvrHealthCheckBuilderExtensions.cs
@@ -79,4 +79,77 @@ public static class IdSvrHealthCheckBuilderExtensions
             tags,
             timeout));
     }
+
+    /// <summary>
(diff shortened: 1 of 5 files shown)


m-gug:
@dotnet-policy-service agree

m-gug:
I have added the possibility that the signature algorithms can be passed as parameters.
It's not the prettiest implementation, but it's backwards compatible.

pbrosl:
Can this PR be merged? We also need for the signature algorithms to be configurable when setting up the Healthchecks

Issue #2261, opened 2024-07-12
Add support for basic authentication to the Solr health check 


HalHunt:
**What would you like to be added**:
I didn't see a way to supply a username and password to the Solr health check.  I tired to add the username and password to the SolrOptions.Uri as in http://solr:SolrRocks@localhost:8983/solr without success.  Am I missing anything?

**Why is this needed**:
To allow the Solr health check to be used in more realistic enterprise environments that use SSL and basic authentication.

The fix is relatively easy and I have a PR in the works however I need some guidance on unit tests.  I have the correct new unit tests but I'm not sure how to support both Solr configurations (i.e. no auth and with auth) in the same docker container.  Any guidance is greatly appreciated.

Alirexaa:
Could you send PR?
I will check it.

HalHunt:
@Alirexaa done!  https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2263

HalHunt:
Found this to configure Solr to support both no auth + basic auth.

`{
  "authentication":{ 
    "blockUnknown": false, 
    "class":"solr.BasicAuthPlugin",
    "credentials":{"solr":"IV0EHq1OnNrj6gvRCwvFwTrZ1+z1oBbnQdiVC3otuq0= Ndd7LKvVBAaZIF0QAVi1ekCfAJXr1GGfLtRUXhgrF8c="}, 
    "realm":"solr", 
    "forwardCredentials": false 
  },
  "authorization":{
    "class":"solr.RuleBasedAuthorizationPlugin",
    "permissions":[{"name":"security-edit",
        "role":"admin"}], 
    "user-role":{"solr":"admin"} 
  }
}`

PR #2263, opened 2024-07-14
2261 Add basic authentication support to Solr health check

HalHunt:
Adds basic authentication support to the Solr health check.

Most enterprise environments configure Solr to use SSL and some sort of authentication.  This PR adds basic authentication support to the Solr health check.

#2261 

Some of the existing Solr unit tests required a higher timeout from the default of 1 second in order to succeed.  However I did not commit those temporary changes since it's likely it's my local testing environment.  Also the new unit test I wrote requires Solr to be configured to use basic authentication.

No.  All changes are opt-in/optional.

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.Solr/DependencyInjection/SolrHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.Solr/DependencyInjection/SolrHealthCheckBuilderExtensions.cs
@@ -34,7 +34,7 @@ public static class SolrHealthCheckBuilderExtensions
         TimeSpan? timeout = default)
     {
         var options = new SolrOptions();
-        options.UseServer(solrUri, solrCore, timeout: null);
+        options.UseServer(solrUri, solrCore, username: null, password: null, timeout: null);
 
         return builder.Add(new HealthCheckRegistration(
             name ?? NAME,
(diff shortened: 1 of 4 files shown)


HalHunt:
@microsoft-github-policy-service agree company="NEOGOV"

dario-axxes:
@Alirexaa Any chance you'll be able to look at this in the near future? Are currently blocked on this.

adamsitnik (commented):
I am supportive of the idea of making Solr health check more configurable, but I wonder if we should just ask the users for a factory method that returns `SolrConnection`.

Something similar to https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/2040, where I've enforced the Azure SDK best practices for most of the Azure health checks.

```cs
public static IHealthChecksBuilder AddDatadogPublisher(
    string serviceCheckName,
    Func<IServiceProvider, SolrConnection>? connectionFactory = default, 
    string[]? defaultTags = default)
{
            name ?? NAME,
            sp => new SolrHealthCheck(connectionFactory?.Invoke() ?? sp.GetRequiredService<SolrConnection>()),
            failureStatus,
            tags,
            timeout));
}
```

Because from what I can see it's own DI package registers it as a singleton:

https://github.com/SolrNet/SolrNet/blob/4955ddba1cce3ca3d5f7a6859dd9adc2d37ac770/Microsoft.DependencyInjection.SolrNet/ServiceCollectionExtensions.cs#L89

If we do that, the users are going to need only configure Solr in one place and the health checks is simply going to re-use that and be unaware of all the config possibilities.

Issue #2232, opened 2024-06-06
Make MongoDB health check a retryable operation


MaxLautsch:
**What would you like to be added**:
Due to issue https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/617 the command to test MongoDBs availability was changed to a db.runCommand({ping: 1}) when a database is specified which resolves that problem, but introduces a condition that could cause intermittent unhealthy statuses.

I work for MongoDB and would be happy to put you in contact with the MongoDB C#/.Net driver team to work on a solution.

**Why is this needed**:
The issue is related to retryable operations ([retryable reads](https://www.mongodb.com/docs/manual/core/retryable-reads/) and [retryable writes](https://www.mongodb.com/docs/manual/core/retryable-writes/) ). Essentially for most operations where it is possible, the MongoDB driver itself will retry exactly once to cover switches in the primary and temporary short term network outages.
Due to the RunCommand being a lower level function, according to the [spec](https://github.com/mongodb/specifications/blob/master/source/run-command/run-command.rst#retryability) for it, it is not retryable and this extends to the ping.

This means that in some cases, the check can return a temporary not-ok status, while actually operations would still succeed without errors.

PR #2276, opened 2024-08-03
Make MongoDB health check a retryable operation

smnsht:
 Make MongoDB health check a retryable operation #2232 

In some cases, the check can return a temporary not-ok status, while actually operations would still succeed without errors.

Please reference the issue this PR will close: #2232 

I am not sure if the project AspNetCore.Diagnostics.HealthChecks maintaners wants to rely on MongoDB C#/.Net driver team in the matter of converting ping command to be retryable, so I would like to propose retry logic implemented in the health check itself.

No

- [*] Code compiles correctly
- [*] Created/updated tests
- [*] Unit tests passing
- [*] End-to-end tests passing


src/HealthChecks.MongoDb/MongoDbHealthCheck.cs
```diff
@@ -46,10 +48,30 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
                 // this you can check a specified database.
                 // Related with issue #43 and #617
 
-                await mongoClient
-                    .GetDatabase(_specifiedDatabase)
-                    .RunCommandAsync(_command, cancellationToken: cancellationToken)
-                    .ConfigureAwait(false);
+                for (int attempt = 1; attempt <= MAX_PING_ATTEMPTS; attempt++)
```
WeihanLi, line 51: Should we consider configure a delay between retry and the max retry count

smnsht: @WeihanLi 

I have been thinking about this.

Seems adding a delay between retries won't give any benefit.

If a connection attempt fails due to networking problems, the operation will take a long time to fail. Practically - could take 20-30 seconds, I guess it depends on Mongo client timeout settings, which we don't know.

Let's examine a simple case: MAX_PING_ATTEMPTS = 3, and Mongo client has a timeout of 30 seconds. In this case single check will take ~90 seconds. A second attempt is made after ~30 seconds, and the last attempt is made after ~1minute.

​If a check failed due to permissions/missing collection, etc... - delay between the checks does not have any benefit.

rohitvipin (approved):

adamsitnik (commented):
> Please reference the issue this PR will close: #2232

Please correct me if I am wrong: Mongo DB driver performs 1 retry for most of the operations, but it does not for the method that our heatlh check is using.

The suggested solution is to add such retry on our side.

> In some cases, the check can return a temporary not-ok status, while actually operations would still succeed without errors.

To me it sounds like "by design" (the health check was for some reason unable to perform the ping), but I would prefer to ask @davidfowl and @damianedwards who architect ASP.NET and should be able to tell us what is the right approach in terms of retries performed by the health checks.

DamianEdwards:
My initial thought is that the health check should use whatever resiliency strategy the app would regularly use so that the health check more accurately represents the app's ability to connect to the relevant resource.

src/HealthChecks.MongoDb/MongoDbHealthCheck.cs
adamsitnik, line 51: Let's add a comment to explain why we are doing that:



```suggestion
                // For most operations where it is possible, the MongoDB driver itself will retry exactly once
                // to cover switches in the primary and temporary short term network outages.
                // Due to the RunCommand being a lower level function, according to the spec (https://github.com/mongodb/specifications/blob/master/source/run-command/run-command.rst#retryability)
                // for it, it is not retryable and this extends to the ping.
                for (int attempt = 1; attempt <= MAX_PING_ATTEMPTS; attempt++)

```

adamsitnik (approved):
LGTM, thank you for your contribution @smnsht !

PR #2289, opened 2024-08-21
Added overload to for StatsdConfig

IlSocio:
Was not possible to customize additional parameters of the Datadog Publisher (eg. the port)

I only added an overload to expose the StatsdConfig, so, the consumers of the package can freely set any parameter supported by Datadog.

### DIFF
--- a/src/HealthChecks.Publisher.Datadog/DependencyInjection/DatadogHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.Publisher.Datadog/DependencyInjection/DatadogHealthCheckBuilderExtensions.cs
@@ -24,15 +24,34 @@ public static class DatadogHealthCheckBuilderExtensions
         string datadogAgentName = "127.0.0.1",
         string[]? defaultTags = default)
     {
+        builder.AddDatadogPublisher(serviceCheckName, new StatsdConfig() { StatsdServerName = datadogAgentName }, defaultTags);
+        return builder;
+    }
+
+    /// <summary>
+    /// Add a health check publisher for Datadog.
+    /// </summary>
+    /// <remarks>
+    /// For each <see cref="HealthReport"/> published a custom service check indicating the health check status (OK - Healthy, WARNING - Degraded, CRITICAL - Unhealthy)
+    /// and a metric indicating the total time the health check took to execute in milliseconds is sent to Datadog.
+    /// </remarks>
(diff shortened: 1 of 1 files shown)


IlSocio:
@dotnet-policy-service agree

adamsitnik (commented):
I am supportive of the idea of exposing a possibility for the users to provide their own instance of `StatsdConfig`, but I wonder if we should just ask the users for a factory method that returns `DogStatsdService`.

Something similar to https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/2040, where I've enforced the Azure SDK best practices for most of the Azure health checks.


```cs
public static IHealthChecksBuilder AddDatadogPublisher(
    string serviceCheckName,
    Func<IServiceProvider, DogStatsdService>? serviceFactory = default, 
    string[]? defaultTags = default)
{
    builder.Services
        .AddSingleton<IHealthCheckPublisher>(sp =>
        {
            DogStatsdService dogStatsdService = serviceFactory?.Invoke(sp) ?? sp.GetRequiredService<DogStatsdService>();

            return new DatadogPublisher(dogStatsdService, serviceCheckName, defaultTags);
        });

    return builder;
}
```

@kevingosse could you please point me to the right person that could help us ensure that this health check is optimal?

kevingosse:
Sorry I'm not sure what you mean by "ensuring this health check is optimal"?

adamsitnik:
> Sorry I'm not sure what you mean by "ensuring this health check is optimal"?

@kevingosse Currently the health check registers a singleton factory which creates a `DogStatsdService` dedicated to the health check when its invoked for the first time:

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/f587a637a1dc9850cce87d441ba173d50679aefc/src/HealthChecks.Publisher.Datadog/DependencyInjection/DatadogHealthCheckBuilderExtensions.cs#L28-L38

I am wondering if it should keep doing that, or just resolve `DogStatsdService` instance from the DI (a lot of similar types are thread safe and recommended to used as a singleton). Just to make sure there is one `DogStatsdService` instance per app (rather than at least two).

kevingosse:
Oh I see. Yes, `DogStatsdService` is meant to be used as a singleton, so I agree using DI would help reducing the risk of creating multiple instances 👍

adamsitnik:
I've sent https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2329 as a follow up

PR #2266, opened 2024-07-26
Update MongoDB.Driver Package Version To 2.28.0

Alirexaa:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Please reference the issue this PR will close: #2265

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

Please make sure you've completed the relevant tasks for this PR, out of the following list:

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing
- [ ] End-to-end tests passing
- [ ] Extended the documentation
- [ ] Provided sample for the feature

PR #2291, opened 2024-08-28
Sign HealthChecks.MongoDb

eerhardt:
We can sign this assembly now because MongoDB.Driver is now signed as of v2.28.


Alirexaa:
I remember that I removed `<SignAssembly>false</SignAssembly>` from `HealthChecks.MongoDb` in https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2266.
I checked this and noticed that I removed `<SignAssembly>false</SignAssembly>` from `HealthChecks.UI` instead of `HealthChecks.MongoDb` and it's wrong.
Should we revert this removal from `HealthChecks.UI` or can we sign `HealthChecks.UI` too?
cc @adamsitnik

adamsitnik:
> Should we revert this removal from `HealthChecks.UI` or can we sign `HealthChecks.UI` too?

I don't have a strong opinion, for now we can keep it signed and see how it goes.

Issue #2300, opened 2024-09-23
SendGrid IServiceProvider Overload


cmeyertons:
**What would you like to be added**:

An overload for `AddSendGrid` that accepts a `Func<IServiceProvider, string>` 

**Why is this needed**:

Resolve the apiKey from .NET configuration / DI using Options pattern (defer to resolution time vs during DI configuration phase).  

Better alignment w/ other libs.

PR #2301, opened 2024-09-23
SendGrid ServiceProvider Overload

cmeyertons:

Adds an `IServiceProvider` overload for configuring SendGrid to allow for ApiKey to be pulled from DI (via options pattern, etc)

Please reference the issue this PR will close: #2300 

Not really, no breakages, just a new overload

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [x] Extended the documentation - n/a?
- [x] Provided sample for the feature - tests are the sample?


cmeyertons:
idk who approved the workflow previously, but thank you 😄 

I have updated the public api txt file

cmeyertons:
can i get this merged?

test/HealthChecks.SendGrid.Tests/DependencyInjection/RegistrationTests.cs
```diff
@@ -19,6 +19,30 @@ public void add_health_check_when_properly_configured()
...
+            .AddSendGrid(sp => sp.GetRequiredService<IOptions<SendGridOptions>>().Value.ApiKey);
+
+        using var serviceProvider = services.BuildServiceProvider();
+        var options = serviceProvider.GetRequiredService<IOptions<HealthCheckServiceOptions>>();
+
+        var registration = options.Value.Registrations.First();
+        var check = registration.Factory(serviceProvider);
+
+        registration.Name.ShouldBe("sendgrid");
+        check.ShouldBeOfType<SendGridHealthCheck>();
```
adamsitnik, line 43: Do I read this correctly that this test is not veryfing if value `my_api_key` has been actually used by the health check?

src/HealthChecks.SendGrid/DependencyInjection/SendGridHealthCheckExtensions.cs
```diff
@@ -31,13 +31,41 @@ public static IHealthChecksBuilder AddSendGrid(
         IEnumerable<string>? tags = default,
         TimeSpan? timeout = default)
     {
-        var registrationName = name ?? NAME;
+        return builder.AddSendGrid(_ => apiKey, name, failureStatus, tags, timeout);
```
adamsitnik, line 34: nit: we are now capturing a closure (so there is a performance) penalty, but it helps to reduce code duplication, so I am fine with that. We may however, change that approach in the future.

adamsitnik (commented):
@cmeyertons thank you for your contribution! Overall it LGTM, but please answer to my test question. Thanks!

test/HealthChecks.SendGrid.Tests/DependencyInjection/RegistrationTests.cs
cmeyertons: i used the same assertion code that was acceptable for the previous test.

adamsitnik, line 43: It would be nice to at least check that given overload was used.

```suggestion
        bool called = false;
        var services = new ServiceCollection();

        services.AddOptions<SendGridOptions>().Configure(options => options.ApiKey = "my_api_key");
        services.AddHealthChecks()
            .AddSendGrid(sp =>
            {
                called = true;
                return sp.GetRequiredService<IOptions<SendGridOptions>>().Value.ApiKey;
            });

        using var serviceProvider = services.BuildServiceProvider();
        var options = serviceProvider.GetRequiredService<IOptions<HealthCheckServiceOptions>>();

        var registration = options.Value.Registrations.First();
        var check = registration.Factory(serviceProvider);

        registration.Name.ShouldBe("sendgrid");
        check.ShouldBeOfType<SendGridHealthCheck>();
        called.ShouldBeTrue();
```

adamsitnik (approved):
LGTM, thank you @cmeyertons !

Issue #2178, opened 2024-03-14
Would be useful if UIResponseWriter had more extensibility


terjew:
**What would you like to be added**:
Some way of overriding or extending the json settings used to serialize the UI Report. Eg. make a new static function taking a `Lazy<JsonSerializerOptions>` as argument and returning a `Func<HttpContext, HealthReport, Task>`.

**Why is this needed**:
`UIResponseWriter` fills almost all of my needs when it comes to giving detailed responses to health checks from my services, even when I am not using the rest of the Health Check UI. However, the fact that it doesn't "pretty-print" the json (writes it indented) makes it slightly less usable, and I couldn't find any way to override the json settings without replacing the entire class.

It's a tiny feature, but it would make it possible to re-use this otherwise very useful functionality in other health check cases.

PR #2307, opened 2024-10-22
#2178 Added a new method CreateResponseWriter to UIResponseWriter

terjew:
Makes it possible for other projects to reuse UIResponseWriter even if they have other needs with regards to serialization options.

#2178 

Do I need to document the functionality? A sample use case can be seen in the unit test.

All the tests from UI.Client and UI.Core run fine, which is what I would expect to be affected by the changes.

Many tests from other projects do not run successfully on my local setup, but that is also true before applying my changes. I assume I am missing some dependency, configuration or similar. It seems like there is a CI workflow to run tests in GitHub, but it is waiting for approval before it will run.

The proposed change should only affect library users, not end-users.

- [X] Code compiles correctly
- [X] Created/updated tests
- [X] Unit tests passing


terjew:
@Alirexaa how do I get a reviewer assigned to this?

Alirexaa:
> @Alirexaa how do I get a reviewer assigned to this?

Sorry. I'm not familiar with UI-related things.

Issue #2186, opened 2024-04-01
Add Qdrant health check client


timheuer:
**What would you like to be added**:
Add support for Qdrant
https://qdrant.tech/documentation/guides/monitoring/?q=health

PR #2310, opened 2024-10-30
Add Qdrant health checks

Alirexaa:

Please reference the issue this PR will close: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/2186

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing


src/HealthChecks.Qdrant/HealthChecks.Qdrant.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <TargetFrameworks>$(DefaultLibraryTargetFrameworks)</TargetFrameworks>
+    <PackageTags>$(PackageTags);Qdrant</PackageTags>
+    <Description>HealthChecks.Qdrant is the health check package for Qdrant.</Description>
+    <VersionPrefix>$(HealthCheckQdrant)</VersionPrefix>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Qdrant.Client" Version="1.12.0" />
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" Version="8.0.0" />
+    <!--<Compile Include="../HealthCheckResultTask.cs" />-->
```
eerhardt, line 13: Can this be removed?

src/HealthChecks.Qdrant/QdrantHealthCheck.cs
```diff
@@ -0,0 +1,31 @@
...
+/// </summary>
+public class QdrantHealthCheck : IHealthCheck
+{
+    private readonly QdrantClient _client;
+
+    public QdrantHealthCheck(QdrantClient? client)
+    {
+        _client = Guard.ThrowIfNull(client);
+    }
+    public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
```
eerhardt, line 17:
```suggestion

    /// <inheritdoc />
    public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
```

src/HealthChecks.Qdrant/QdrantHealthCheck.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace HealthChecks.Qdrant;
+
+/// <summary>
+/// A health check for Qdrant services.
+/// </summary>
+public class QdrantHealthCheck : IHealthCheck
+{
+    private readonly QdrantClient _client;
+
+    public QdrantHealthCheck(QdrantClient? client)
```
eerhardt, line 13:
```suggestion
    public QdrantHealthCheck(QdrantClient client)
```

This shouldn't be nullable if we are going to throw when it is null.

src/HealthChecks.Qdrant/QdrantHealthCheck.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        _client = Guard.ThrowIfNull(client);
+    }
+    public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            var response = await _client.HealthAsync(cancellationToken).ConfigureAwait(false);
+            return !(response is null || response.Title is null)
```
eerhardt, line 22:
```suggestion
            return response?.Title is not null
```

test/HealthChecks.Qdrant.Tests/Functional/QdrantHealthCheckTests.cs
```diff
@@ -0,0 +1,85 @@
...
+                {
+                    Predicate = r => r.Tags.Contains("qdrant")
+                });
+            });
+
+        using var server = new TestServer(webHostBuilder);
+
+        using var response = await server.CreateRequest("/health").GetAsync();
+
+        response.StatusCode.ShouldBe(HttpStatusCode.OK);
```
eerhardt, line 31: For my learning, how does this return OK when talking to the localhost? There isn't a Qdrant instance running, is there?

Alirexaa: Qdrant instance running by ci action
see: 
https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2310/files#diff-5d6b2fd45c59f6628e08c23dcf560e4b79d0a6484e47de1a258a58d063d2f21aR31-R36

In most cases, we use this approach to run services in this repo.

eerhardt: But if I just run the test from Test Explorer, it fails, right?

eerhardt (approved):
This looks great to me. Thanks for pushing this through.

@adamsitnik @unaizorrilla - any thoughts?

test/HealthChecks.Qdrant.Tests/Functional/QdrantHealthCheckTests.cs
Alirexaa: I think so.

test/HealthChecks.Qdrant.Tests/DependencyInjection/RegistrationTests.cs
```diff
@@ -0,0 +1,64 @@
...
+        var options = serviceProvider.GetRequiredService<IOptions<HealthCheckServiceOptions>>();
+
+        var registration = options.Value.Registrations.First();
+        var check = registration.Factory(serviceProvider);
+
+        registration.Name.ShouldBe("qdrant");
+        check.ShouldBeOfType<QdrantHealthCheck>();
+    }
+
+    private QdrantClient _clientFactory(IServiceProvider _)
```
adamsitnik, line 62: I see what you are trying to achieve, but I believe that using an underscore for the name of a private method is a bad idea, as most of the C# devs would expect it to be a field.

src/HealthChecks.Qdrant/QdrantHealthCheck.cs
```diff
@@ -0,0 +1,32 @@
...
+        _client = Guard.ThrowIfNull(client);
+    }
+
+    /// <inheritdoc />
+    public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            await _client.HealthAsync(cancellationToken).ConfigureAwait(false);
+
```
adamsitnik, line 24:
```suggestion

```

src/HealthChecks.Qdrant/QdrantHealthCheck.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// <inheritdoc />
+    public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            await _client.HealthAsync(cancellationToken).ConfigureAwait(false);
+            return HealthCheckResult.Healthy();
+        }
+        catch (RpcException ex)
```
adamsitnik, line 26:
```suggestion
        catch (Exception ex)
```

PR #2314, opened 2024-11-11
Update healthchecksui.css (background color on logo element)

brgrz:
Background color shouldn't be used for logo image because it conflicts with images that use transparency for example SVGs (which are the standard format for logos on the web nowadays)


brgrz:
@dotnet-policy-service agree

PR #2323, opened 2024-11-25
Add support for RabbitMQ.Client version 7.

### LINKED ISSUE #2319: Upgrade RabbitMQ.Client to v7
To upgrade HealthChecks.RabbitMQ from v6 to v7.

**Why is this needed**:
Because v7 has lots of API changes in compare to v6, HealthChecks.RabbitMQ blocks users from upgrade the RabbitMQ client in their apps.

**More info**
- Release notes: https://github.com/rabbitmq/rabbitmq-dotnet-client/releases/tag/v7.0.0
- Migration guide: https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/main/v7-MIGRATION.md
- Discussion topic about migration: https://github.com/rabbitmq/rabbitmq-dotnet-client/discussions/1720

eerhardt:
RabbitMQ.Client version 7 made major breaking changes - interface renames, all sync methods are removed and only async methods remain.

Handle this breaking change by splitting our package into 2, one for each major version.

1. For the current HealthChecks.Rabbitmq package, update to the new 7.0.0 version and update so the health checks will work with v7.
2. We add a new, forked component named HealthChecks.Rabbitmq.v6 which will have a dependency on 6.x . People who want to keep using the version 6 can opt into using this package. We put a NuGet version limit on our dependency: [6.8.1,7.0.0). This way people won't be able to update to the 7.0.0 version, which will break their app. They will need to migrate back to the base package to use version 7.
3. When RabbitMQ.Client v6 is no longer supported, we can dead-end the `HealthChecks.Rabbitmq.v6` package.

Fix #2319

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing


test/HealthChecks.RabbitMQ.v7.Tests/HealthChecks.RabbitMQ.v7.Tests.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <AssemblyName>HealthChecks.RabbitMQ.Tests</AssemblyName>
+  </PropertyGroup>
+  
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\HealthChecks.Rabbitmq.v7\HealthChecks.Rabbitmq.v7.csproj" />
+
+    <Compile Include="../HealthChecks.RabbitMQ.Tests/DependencyInjection/RegistrationTests.cs" Link="DependencyInjection/RegistrationTests.cs" />
+    <Compile Include="../HealthChecks.RabbitMQ.Tests/Functional/RabbitHealthCheckTests.cs" Link="Functional/RabbitHealthCheckTests.cs" />
+
+    <None Include="../HealthChecks.RabbitMQ.Tests/HealthChecks.Rabbitmq.approved.txt" />
```
eerhardt, line 13:
```suggestion
```

This can be removed.

adamsitnik (commented):
We have not released the 9.0 version yet, so it theory we could include this breaking change now and then release 9.0. So those who need the old packages would keep using the 8.0 version.

Thoughts?

AnthonyDewhirst:
When is the planned release date of release 9.0?
We use RabbitMQ for local testing, so I am comfortable commenting out the health check for now, but ideally would like to not do this. As mentioned 8 doesn't work when upgrading other parts to .net 9

Alirexaa (commented):
Tests only run for HealthChecks.Rabbitmq. Tests should be run for HealthChecks.Rabbitmq.v7 too.

src/HealthChecks.Rabbitmq/RabbitMQHealthCheck.cs
```diff
@@ -26,49 +26,88 @@ public RabbitMQHealthCheck(RabbitMQHealthCheckOptions options)
...
+            }
 
-                return factory.CreateConnection();
-            });
+            return await factory.CreateConnectionAsync(cancellationToken).ConfigureAwait(false);
+        }).ConfigureAwait(false);
+}
 
-        return _connection;
+internal static class ConcurrentDictionaryExtensions
```
adamsitnik, line 73: It's not related to this PR, but it shows that we really need to get rid of that static cache, as it brings so much complexity and increases a risk for various bugs (leaks etc).

adamsitnik (approved):
LGTM, big thanks for putting so much attention into details @eerhardt !

src/HealthChecks.Rabbitmq/RabbitMQHealthCheck.cs
eerhardt: Yes - agreed. I will try to make that change next.

PR #2324, opened 2024-11-25
Add support for MongoDB.Driver version 3

### LINKED ISSUE #2322: Compatibility issue with MongoDB.Driver 3.0.0
I have issue with last mongoDB.Driver 3.0.0

System.IO.FileNotFoundException: Could not load file or assembly 'MongoDB.Driver.Core, Version=2.28.0.0, Culture=neutral, PublicKeyToken=94992a530f44e321'. The system cannot find the file specified.
2024-11-24T18:30:59.8559813Z
2024-11-24T18:30:59.8559841Z       File name: 'MongoDB.Driver.Core, Version=2.28.0.0, Culture=neutral, PublicKeyToken=94992a530f44e321'
2024-11-24T18:30:59.8563412Z          at HealthChecks.MongoDb.MongoDbHealthCheck.CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken)
2024-11-24T18:30:59.8563494Z          at Microsoft.Extensions.Diagnostics.HealthChecks.DefaultHealthCheckService.RunCheckAsync(HealthCheckRegistration registration, CancellationToken cancellationToken)

eerhardt:
MongoDB.Driver version 3 made a major breaking change by removing the MongoDB.Driver.Core assembly.

~Handle this breaking change by splitting our package into 2, one for each major version.~

~1. For the current HealthChecks.MongoDb package, we put a NuGet version limit on our dependency: [2.28.0,3.0.0). This way people won't be able to update to the 3.0.0 version, which will break their app.~
~2. We add a new, forked component named HealthChecks.MongoDb.v3 which will have a dependency on 3.0.0 and contains updates so the health checks will work with v3. People who explicitly want to use version 3 can opt into using this package.~
~3. When the next major version of HealthChecks ships, we can "swap" the dependencies around. The HealthChecks.MongoDb package will be updated to depend on version 3 of MongoDB.Driver. If MongoDB.Driver v2.x is still in support, we can create HeatlhChecks.MongoDb.v2 which has the dependency limit [2.28.0,3.0.0) and works with the version 2 of MongoDB.Driver. HealthChecks.MongoDb.v3 will be dead-ended.~

~Similar changes as [Add support for RabbitMQ.Client version 7. (Xabaril/AspNetCore.Diagnostics.HealthChecks#2323)](https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2323).~

Handle this breaking change by updating to the new v3 version in HealthChecks `v9.x` release. According to https://www.mongodb.com/community/forums/t/net-driver-2-30-0-released/301201, MongoDB.Driver `v2.x` will no longer be supported:

> Version 2.30.0 of the driver is the last scheduled 2.x release. Further driver development will continue in 3.x versions.

Because of this, the HealthChecks should only need to support v3+ going forward. If someone needs to use the `v2.x` version, they can continue to use AspNetCore.HealthChecks.MongoDb `v8.x`.

Fix #2322

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing


rohitvipin (approved):

src/HealthChecks.MongoDb/HealthChecks.MongoDb.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="MongoDB.Driver" Version="2.28.0" />
+    <PackageReference Include="MongoDB.Driver" Version="[2.28.0,3.0.0)" />
```
rohitvipin, line 11: Should this be 2.30.0?

src/CallerArgumentExpressionAttribute.cs
```diff
@@ -21,7 +21,7 @@ public CallerArgumentExpressionAttribute(string parameterName)
 
 #endif
 
-#if NETSTANDARD2_0
+#if NETSTANDARD2_0 || NETFRAMEWORK
```
adamsitnik, line 24: Should we change the condition to also include NS2.1?

```suggestion
#if NETSTANDARD || NETFRAMEWORK
```

src/HealthChecks.MongoDb/HealthChecks.MongoDb.csproj
```diff
@@ -1,7 +1,8 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>$(DefaultLibraryTargetFrameworks)</TargetFrameworks>
+    <!-- Match the TFMs of MongoDB.Driver -->
+    <TargetFrameworks>$(DefaultNetCoreApp);netstandard2.1;net472</TargetFrameworks>
```
adamsitnik, line 5: Perhaps we should simply drop the NS support?

```suggestion
    <TargetFrameworks>$(DefaultNetCoreApp);net472</TargetFrameworks>

```

eerhardt: Since the underlying `MongoDB.Driver` library supports `netstandard2.1`, I feel like the health check library should as well. It would be a pity to block someone for no reason.

For example, I think Unity supports netstandard2.1 - https://docs.unity3d.com/6000.0/Documentation/Manual/dotnet-profile-support.html

src/CallerArgumentExpressionAttribute.cs
eerhardt: No, the `NotNullAttribute` is in netstandard2.1, just not netstandard2.0 (or netfx).

https://apisof.net/catalog/c9927a20ea8a7ffe46c456b1d55b5d70

src/HealthChecks.MongoDb/HealthChecks.MongoDb.csproj
eerhardt: Check the top comment. This is outdated now. We will only support the 3+ version going forward.

eerhardt:
@adamsitnik - I believe this is ready for re-review. I have addressed all the feedback.

adamsitnik (approved):
LGTM, thank you @eerhardt !

Kiechlus:
Thanks a lot everyone! @adamsitnik, is it possible to estimate when this will be available as [nuget](https://www.nuget.org/packages/AspNetCore.HealthChecks.MongoDb) package. Or can we somehow get the preview package?

Background, we have a release coming and need to get our MongoDb health checks green, the one way or the other.

adamsitnik:
> is it possible to estimate when this will be available as [nuget](https://www.nuget.org/packages/AspNetCore.HealthChecks.MongoDb) package

I am aiming for next week, but it depends on the availability of @unaizorrilla

Issue #2325, opened 2024-11-29
HealthChecks.DocumentDb


adamsitnik:
HealthChecks.DocumentDb relies on deprecated Microsoft.Azure.DocumentDB.Core

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/0b1b657dac590cab673787cb3c8bd282388c4574/src/HealthChecks.DocumentDb/HealthChecks.DocumentDb.csproj#L12

From https://www.nuget.org/packages/Microsoft.Azure.DocumentDB.Core:

> Please note, this package is obsolete as of 3/31/2023 and is no longer maintained or monitored. Microsoft encourages you to upgrade to the replacement package, Microsoft.Azure.Cosmos, to continue receiving updates.

We already provide a package that uses the recommended package:

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/tree/master/src/HealthChecks.CosmosDb

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/0b1b657dac590cab673787cb3c8bd282388c4574/src/HealthChecks.CosmosDb/HealthChecks.CosmosDb.csproj#L11

I believe that we should also deprecate [HealthChecks.DocumentDb](https://www.nuget.org/packages/AspNetCore.HealthChecks.DocumentDb/) and remove the source code from the repo.

@eerhardt @unaizorrilla thoughts?

eerhardt:
Sounds good to me.

PR #2327, opened 2024-11-29
Remove vulnerable dependencies

adamsitnik:
I wanted to start preparing for the 9.0 release and when I tried to build the repo I've run into few hundreds of warnings about vulnerable dependencies. I've fixed all of them.

Most of the changes are self-explanatory, for others I've added comments. This PR also includes a fix for #2325

@eerhardt I wonder if we should release a new 8.0.x version after this PR gets merged, before we ship the 9.0 version (.NET 8 is LTS so I expect plenty of people to still use it)


eerhardt:
> I wonder if we should release a new 8.0.x version after this PR gets merged, before we ship the 9.0 version (.NET 8 is LTS so I expect plenty of people to still use it)

The support approach we are taking in .NET Aspire and with the libraries in dotnet/extensions is to use "tip support": https://dotnet.microsoft.com/en-us/platform/support/policy/aspire

> At any time, only the latest release of .NET Aspire is supported by Microsoft. For example: support for .NET Aspire 8.0 ends when .NET Aspire 8.1 is released.

So even if someone is on .NET 8 (LTS), they would use .NET Aspire 9, which supports .NET 8.

We could follow a similar concept here and ship HealthChecks version 9, and support .NET 8.

eerhardt (approved):
LGTM. Thanks!

PR #2329, opened 2024-12-03
Improve Datadog health check

adamsitnik:
1. Ensure we resolve a singleton instance rather than create a dedicated one by default (https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2289#issuecomment-2511930940)
2. Add Readme
3. Update tests so they pass

I've removed the overload that takes a string, so it's a breaking change. It will be released as part of the 9.0 release (major version update).

cc @kevingosse

### DIFF
--- a/src/HealthChecks.Publisher.Datadog/DependencyInjection/DatadogHealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.Publisher.Datadog/DependencyInjection/DatadogHealthCheckBuilderExtensions.cs
@@ -24,15 +24,34 @@ public static class DatadogHealthCheckBuilderExtensions
         string datadogAgentName = "127.0.0.1",
         string[]? defaultTags = default)
     {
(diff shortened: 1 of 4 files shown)


kevingosse (approved):

IlSocio:
due to this issue on Datadog SDK:  https://github.com/DataDog/dogstatsd-csharp-client/issues/91 
(opened since 2019)
the ServiceCheck doesn't honor the value configured within the StasdConfig.Prefix 
(and probably they will not introduce it to avoid breaking compatibility of existing consumers of the API)


So, I'm wondering... 
would you accept a PR to fix this issue on your DatadogPublisher implementation, so that we fix the behaviour for this publisher implementation?

adamsitnik:
TBH I would prefer the issue to be fixed within the client so we don't need to do anything on our side.

PR #2331, opened 2024-12-03
Update to currently supported Target Frameworks.

eerhardt:

This drops support for end-of-life TFMs: anything before `net8.0` keeping `netstandard2.0` in any library that supported it before.

Tests are run on net8 and net9.

Yes, it drops support for end-of-life TFMs in preparation for releasing the next major version.

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -10,7 +10,7 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <Nullable>enable</Nullable>
-    <NoWarn>$(NoWarn);1591;IDISP013;AD0001;</NoWarn> <!--TODO: temporary solution, remove AD0001 after https://github.com/dotnet/aspnetcore/issues/50836 fixed-->
+    <NoWarn>$(NoWarn);1591;IDISP013;</NoWarn>
     <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
   </PropertyGroup>
 
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -20,7 +20,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <DefaultNetCoreApp>net6.0</DefaultNetCoreApp>
+    <DefaultNetCoreApp>net8.0</DefaultNetCoreApp>
     <DefaultLibraryTargetFrameworks>$(DefaultNetCoreApp);netstandard2.0</DefaultLibraryTargetFrameworks>
   </PropertyGroup>
   
--- a/test/HealthChecks.Elasticsearch.Tests/HealthChecks.Elasticsearch.Tests.csproj
+++ b/test/HealthChecks.Elasticsearch.Tests/HealthChecks.Elasticsearch.Tests.csproj
@@ -1,5 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
+  <PropertyGroup>
+    <TestTfmsInParallel>false</TestTfmsInParallel>
+  </PropertyGroup>
+
   <ItemGroup>
     <ProjectReference Include="..\..\src\HealthChecks.Elasticsearch\HealthChecks.Elasticsearch.csproj" />
     <PackageReference Include="Ductus.FluentDocker" />
--- a/test/HealthChecks.Elasticsearch.Tests/HealthChecks.Elasticsearch.Tests.csproj
+++ b/test/HealthChecks.Elasticsearch.Tests/HealthChecks.Elasticsearch.Tests.csproj
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <!-- The functional tests can't be run in parallel because they conflict with each other when creating docker containers. -->
     <TestTfmsInParallel>false</TestTfmsInParallel>
   </PropertyGroup>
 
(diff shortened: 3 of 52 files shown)


src/Directory.Build.props
```diff
@@ -20,7 +20,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <DefaultNetCoreApp>net6.0</DefaultNetCoreApp>
+    <DefaultNetCoreApp>net8.0</DefaultNetCoreApp>
```
Alirexaa, line 23:
```suggestion
    <DefaultNetCoreApp>net8.0;net9.0</DefaultNetCoreApp>
```
Why we don't target `net9` too?

Alirexaa (commented):
Why do tests have target `net9` and `net8`, but libraries only have `net8`?

Directory.Build.props
```diff
@@ -10,7 +10,7 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <Nullable>enable</Nullable>
-    <NoWarn>$(NoWarn);1591;IDISP013;AD0001;</NoWarn> <!--TODO: temporary solution, remove AD0001 after https://github.com/dotnet/aspnetcore/issues/50836 fixed-->
```
adamsitnik, line 13: thank you for removing the outdated TODO 👍

test/HealthChecks.Elasticsearch.Tests/HealthChecks.Elasticsearch.Tests.csproj
```diff
@@ -1,5 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
+  <PropertyGroup>
+    <TestTfmsInParallel>false</TestTfmsInParallel>
```
adamsitnik, line 4: just out of curiosity: why was this required?

adamsitnik (approved):
LGTM, thank you @eerhardt !

src/Directory.Build.props
eerhardt: There's no need to target `net9.0` in a library when it targets `net8.0`, unless you need to use APIs that were introduced in .NET 9. Adding multiple targets just increases the NuGet package size for no benefit.

We follow the same approach in .NET Aspire. Most libraries target `net8.0`, and only if the library needs to use a .NET 9 API (right now only EF Core libraries need this) does it multi-target to both.

The tests, on the other hand, are executable. Multi-targeting them means that the tests will execute on both .NET 8 and .NET 9.

Alirexaa (approved):

test/HealthChecks.Elasticsearch.Tests/HealthChecks.Elasticsearch.Tests.csproj
eerhardt: .NET 9 made a change to run tests in parallel by default: https://learn.microsoft.com/dotnet/core/whats-new/dotnet-9/sdk#run-tests-in-parallel.

This caused these tests to fail because when both the net8 and net9 tests are running at the same time, they both want to create a docker container with the same name:

```
error TESTERROR:
      HealthChecks.Elasticsearch.Tests.Functional.ElasticsearchAuthenticationTests.be_healthy_if_elasticsearch_is_using_valid_user_and_password (1ms):
       Error Message: Class fixture type 'HealthChecks.Elasticsearch.Tests.Fixtures.ElasticContainerFixture' threw in its constructor
      ---- Ductus.FluentDocker.Common.FluentDockerException : Could not create composite service with file(s) D:\git\AspNetCore.Diagnostics.HealthChec
      ks\test\HealthChecks.Elasticsearch.Tests\bin\Debug\net9.0/Resources/docker-compose.yml - result:  Network resources_default  Creating
       Network resources_default  Created
       Volume "resources_certs"  Creating
       Volume "resources_certs"  Created
       Volume "resources_esdata01"  Creating
       Volume "resources_esdata01"  Created
       Container resources-setup-1  Creating
      Error response from daemon: Conflict. The container name "/resources-setup-1" is already in use by container "aae4ab5800169ebc71eab2f914e418da6f
      156e283f373d34b44b431805fccca5". You have to remove (or rename) that container to be able to reuse that name.
      Stack Trace:

      ----- Inner Stack Trace -----
         at Ductus.FluentDocker.Services.Impl.DockerComposeCompositeService.Start()
         at Ductus.FluentDocker.Services.Impl.DockerComposeCompositeService.Ductus.FluentDocker.Services.ICompositeService.Start()
         at HealthChecks.Elasticsearch.Tests.Fixtures.ElasticContainerFixture..ctor() in D:\git\AspNetCore.Diagnostics.HealthChecks\test\HealthChecks.
      Elasticsearch.Tests\Fixtures\ElasticContainerFixture.cs:line 26
         at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
         at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

Telling `dotnet test` to run each TFM targetted test serially fixes this issue. I'll add a quick comment.

Alirexaa: Not related to this PR but i think we should change the tests to be able run on local. Currently tests only work when they run in CI.

eerhardt: Some tests run locally. Some fail because the external container isn't set up - which you need to do manually.

I agree - it would be nice to be able to just run the tests locally. We could consider using TestContainers in a fixture, like we do in .NET Aspire.

Issue #2087, opened 2023-11-03
Dynamic Connection String Not Refreshed in Subsequent Health Checks


MGithubber:
**What happened:**
The AddNpgSql extension method for health checks caches the first NpgsqlDataSource created by the dbDataSourceFactory, which prevents the factory from being invoked on subsequent health checks. This is problematic for scenarios where the database connection string changes over time.

**What you expected to happen:**
I expected the dbDataSourceFactory to be called on every health check invocation to ensure that the most current connection string is used.

**How to reproduce it**:
1) Configure the health checks using the AddNpgSql method with a connectionStringFactory that provides a dynamic connection string.
2) Call the health check endpoint once.
3) Change the underlying connection string that the connectionStringFactory provides.
4) Run the health check again and notice that the change in connection string is not reflected in the health check.

**Source code sample**:
```csharp
using HealthChecks.UI.Client;

using Microsoft.AspNetCore.Diagnostics.HealthChecks;
using Microsoft.Extensions.Options;

var builder = WebApplication.CreateBuilder(args);


builder.Services
    .AddOptions<ConnectionStringOptions>()
    .BindConfiguration(ConnectionStringOptions.SectionName);

builder.Services.AddHealthChecks() // The lambda retrieves the current DefaultDb connection string from the configuration.
    .AddNpgSql(services => services.GetRequiredService<IOptionsSnapshot<ConnectionStringOptions>>().Value.DefaultDb);


var app = builder.Build();


app.UseHttpsRedirection();

app.MapHealthChecks("_health", new HealthCheckOptions
{
    ResponseWriter = UIResponseWriter.WriteHealthCheckUIResponse
});

app.Run();



public class ConnectionStringOptions
{
    public const string SectionName = "ConnectionStrings";

    public string DefaultDb { get; set; } = default!;
}
```

**Environment**:
- .NET Core version: .NET 7
- Healthchecks version: 7.1.0
- Operative system: Windows 11 Pro

sungam3r:
@adamsitnik could you please give a comment? I know you've refactored npgsql HC package.

adamsitnik:
I've introduced this bug in 7.1.0 and I am afraid that it's still a thing with 8.0.0.

The options instance is cached by capturing it in a closure;

[...]

PR #2334, opened 2024-12-05
Remove private static cache from MongoDbHealthCheck

### LINKED ISSUE #2148: Remove private static cache from MongoDbHealthCheck
`MongoDbHealthCheck` is maintaining it's private static cache of instances. It's wrong, as it can cause a memory leak (it won't ever be freed).

Moreover, it's can create instances of `MongoClient`. This is wrong, as it can lead into a situation when we have multiple instances of `MongoClient` that are connected to the same DB: one used by the app and another created and used by the health check. And we should have only one.

We should do the same as in #2040, #2116 and #2096

adamsitnik:
fixes #2148


Alirexaa (approved):
LGTM

adamsitnik:
@eerhardt this change will force the Aspire to do the right thing as well, as now it's using the overload that provides a connection string (and creates a new instance of the client just for the health check)

https://github.com/dotnet/aspire/blob/main/src/Components/Aspire.MongoDB.Driver/AspireMongoDBDriverExtensions.cs#L171

src/HealthChecks.MongoDb/DependencyInjection/MongoDbHealthCheckBuilderExtensions.cs
```diff
@@ -12,40 +12,14 @@ public static class MongoDbHealthCheckBuilderExtensions
     private const string NAME = "mongodb";
 
     /// <summary>
-    /// Add a health check for MongoDb database that list all databases on the system.
+    /// Add a health check for MongoDb that list all databases from specified <paramref name="clientFactory"/>.
```
eerhardt, line 15: This summary is incorrect in some cases because if you pass a databaseName it will just ping that database.

src/HealthChecks.MongoDb/DependencyInjection/MongoDbHealthCheckBuilderExtensions.cs
```diff
@@ -56,198 +30,27 @@ public static IHealthChecksBuilder AddMongoDb(
     /// <returns>The specified <paramref name="builder"/>.</returns>
     public static IHealthChecksBuilder AddMongoDb(
         this IHealthChecksBuilder builder,
-        string mongodbConnectionString,
-        string mongoDatabaseName,
+        Func<IServiceProvider, IMongoClient>? clientFactory = default,
+        Func<IServiceProvider, string>? databaseNameFactory = default,
```
eerhardt, line 34: Should we also have an overload that instead returns an `IMongoDatabase` instead of both of these?

adamsitnik: Good point. I've added such an overload, but I could not have made the factory optional argument, as compiler would not be able to choose what `.AddMongo()` means. I expect `IMongoDatabase` to be less frequently used, so I've made the `Func<IServiceProvider, IMongoDatabase> dbFactory` argument mandatory.

eerhardt: > so I've made the Func<IServiceProvider, IMongoDatabase> dbFactory argument mandatory.

Yep. that sounds right to me.

eerhardt (approved):
LGTM. Thanks!

src/HealthChecks.MongoDb/DependencyInjection/MongoDbHealthCheckBuilderExtensions.cs
```diff
@@ -56,56 +30,35 @@ public static IHealthChecksBuilder AddMongoDb(
...
         string? name = default,
         HealthStatus? failureStatus = default,
         IEnumerable<string>? tags = default,
         TimeSpan? timeout = default)
     {
         return builder.Add(new HealthCheckRegistration(
             name ?? NAME,
-            sp => new MongoDbHealthCheck(mongodbConnectionString, mongoDatabaseName),
+            sp => Factory(sp, clientFactory, databaseNameFactory),
```
eerhardt, line 42: @adamsitnik - this will execute the `clientFactory` every time health checks are run. Is that intentional?

adamsitnik: @eerhardt yes, because some clients may register something else than a singleton. For example a client instance that is valid only for some period of time like in https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/2087

PR #2336, opened 2024-12-05
Use recommended NATS.Net client for Nats package

### LINKED ISSUE #2199: Support Nats authentication using Username and Password
Support the Username and Password authentication.
https://docs.nats.io/using-nats/developer/connecting/userpass

**Why is this needed**:
It would be beneficial to support more authentication methods since they are available in Nats.

Alirexaa:
This pull request introduces several significant changes to the NATS health check implementation. The updates include replacing the `NATS.Client` package with `NATS.Net`, refactoring the `NatsHealthCheck` class to use the new client, and updating the health check registration and tests accordingly.

### Dependency updates:
* Replaced `NATS.Client` package with `NATS.Net` package in project files and updated the relevant using directives. [[1]](diffhunk://#diff-5baf5f9e448ad54ab25a091adee0da05d4d228481c9200518fcb1b53a65d4156L74-R74) [[2]](diffhunk://#diff-61050339142660b578067c745df4775909c98343bc9d4b13f6f6bfb58dcefceeL11-R11) [[3]](diffhunk://#diff-3489fdbbfb4c70bc5cf3fffe90c99d6bfe6f24b7ecbcd5c3df982becf31da140R3) [[4]](diffhunk://#diff-8ac3efba3d09b5ef082fc4ab28e3a1ce3a0057bef34fe6bf10015ef9aa565ff7L1-L110)

### Code refactoring:
* Refactored `NatsHealthCheck` class to use `INatsConnection` from `NATS.Client.Core` and simplified the health check logic.
* Removed `NatsOptions` class, as it is no longer needed with the new `NATS.Net` client.

### Health check registration:
* Updated `NatsHealthCheckBuilderExtensions` to support a client factory for creating `INatsConnection` instances and simplified the method signatures.

### Tests:
* Updated unit tests and functional tests to use the new `NATS.Net` client and modified the test cases to reflect the changes in the health check registration and implementation. [[1]](diffhunk://#diff-4496faf29256b0f312b891e527c2d9511c121d174c476ae5785ab70bcf060f2aR1-R48) [[2]](diffhunk://#diff-07eb35e4889ba7b1b585187f458c51758a687fc172563276cb39ed94caf7e2b5L2-L124)

It's recommended to use NATS.Net for  .NET 6 upwards.
Also, we use the same approach (https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/2040) to register client singleton.

Also, this PR should fix #2199 

Yes. `NATS.Client` removed. All previews `AddNats` were removed and a new one using the factory method was introduced.

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [ ] Extended the documentation (Readme should be updated after approval)


src/HealthChecks.Nats/NatsHealthCheck.cs
```diff
@@ -1,111 +1,36 @@
...
-
-    private readonly NatsOptions _options;
-
-    private IConnection? _connection;
-
-    public NatsHealthCheck(NatsOptions natsOptions)
+    public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
     {
-        _options = Guard.ThrowIfNull(natsOptions);
+        var result = connection.ConnectionState switch
```
adamsitnik, line 13: How about just always calling `ConnectAsync()`?

It will achieve the same effect:
- It will just return a finished task for `Open` connection
- try to re-open the closed connections which should help us avoid getting into issues like https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/issues/1544

But also wait for the reconnection to finish for `Connecting` and `Reconnecting`. In case the enum is ever extended with a new value, `ConnectAsync()` will most likely handle it for us.

https://github.com/nats-io/nats.net/blob/3f919ddcd55ea0f765b4ce9b4137e2a58c589683/src/NATS.Client.Core/NatsConnection.cs#L155-L183

test/HealthChecks.Nats.Tests/Functional/NatsHealthCheckTests.cs
```diff
@@ -1,125 +1,105 @@
...
-
-    [Fact]
-    public Task be_unhealthy_if_nats_endpoint_is_bogus() =>
-        FactAsync(
-            setup => setup.Url = "bogus",
-            async response =>
+                services
+                    .AddHealthChecks()
+                    .AddNats(
+                        clientFactory: sp =>
```
adamsitnik, line 18: nit: some of our users may be searching for the samples in our test project. This code is valid, but it registers a factory that is going to create a new connection instance every time a health check is going to be invoked (which may lead to resource leaks etc).

Would you have some time to use NATS.Extensions.Microsoft.DependencyInjection instead and just call `AddNatsClient` from that package?

Here I've found some samples (in a test project ;)): https://github.com/nats-io/nats.net/blob/3f919ddcd55ea0f765b4ce9b4137e2a58c589683/tests/NATS.Extensions.Microsoft.DependencyInjection.Tests/NatsHostingExtensionsTests.cs#L73

adamsitnik (commented):
The changes look good, I've added some ideas for minor improvements. Thank you @Alirexaa !

BTW is there any chance you could find some time to add a `README.md` and use `NATS.Extensions.Microsoft.DependencyInjection` in the code sample? We do sth similar for Npgsql: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/tree/master/src/HealthChecks.NpgSql#recommended-approach

And it would be nice to describe the breaking change in the README as well

test/HealthChecks.Nats.Tests/Functional/NatsHealthCheckTests.cs
Alirexaa: I did not use `AddNatsClient` in this test because I wanted to test the client factory.
But you are right. It may lead to resource leaks. So, I kept the client factory but changed it to return the created client, not run a new one.

src/HealthChecks.Nats/NatsHealthCheck.cs
```diff
@@ -1,111 +1,38 @@
...
-/// </remarks>
-public sealed class NatsHealthCheck : IHealthCheck, IDisposable
+public sealed class NatsHealthCheck(INatsConnection connection) : IHealthCheck
 {
-    private static readonly ConnectionFactory _connectionFactory = new();
-
-    private readonly NatsOptions _options;
+    public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+    {
+        await TryConnectAsync(connection).ConfigureAwait(false);
```
adamsitnik, line 13: I meant just calling `ConnectAsync`. If it's in a healthy state, it just returns `Healthy`, if not, it throws and we return `Unhealthy`.

```suggestion
        return await TryConnectAsync(connection).ConfigureAwait(false);
```

adamsitnik (approved):
LGTM, thank you very much for your contribution @Alirexaa !

Issue #2298, opened 2024-09-16
Allow registering multiple Kafka health checks.


mitchdenny:
**What would you like to be added**:

Currently the `AddKafka` results in a singleton of the KafkaHealthCheckOptions being registered with DI:

https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/116a4eac95eac47d1811d33a1f93b053ff87fa19/src/HealthChecks.Kafka/DependencyInjection/KafkaHealthCheckBuilderExtensions.cs#L38

This should be adjusted to be more like the other health checks so that multiple health checks can be registered.

**Why is this needed**:

If you register multiple health checks they all end up having the same options which isn't very useful. I ran into this whilst adding support for Kafka health checks in the .NET Aspire application model and I was scratching my head as to why one of my Kafka health checks wasn't returning healthy - until I figured out it was looking at the write Kafka instance (which wasn't running yet). See:

https://github.com/dotnet/aspire/pull/5719

PR #2337, opened 2024-12-05
Add an overload to Kafka extensions for registering multiple instances

Alirexaa:
This pull request introduces a new overload for the `AddKafka` method in the `KafkaHealthCheckBuilderExtensions` class, which allows configuring Kafka health checks using an `optionFactory`. Additionally, the corresponding method signature is added to the approved API file.

### New Overload for Kafka Health Check Configuration:

* [`src/HealthChecks.Kafka/DependencyInjection/KafkaHealthCheckBuilderExtensions.cs`](diffhunk://#diff-53a9d856538709a70baba1b85a3c3e09ddec83b83425e1f25d835acd80986b78R116-R147): Added a new overload of the `AddKafka` method that accepts an `optionFactory` to configure Kafka health checks. This method includes parameters for the health check name, failure status, tags, and timeout.

### Update to Approved API:

* [`test/HealthChecks.Kafka.Tests/HealthChecks.Kafka.approved.txt`](diffhunk://#diff-5554acf5cd4192c66d18f15779d046193ce60c9b4fbc5bd0e00010bd14fae824R23): Added the new overload of the `AddKafka` method to the approved API file to reflect the changes in the method signature.

Please reference the issue this PR will close: #2298 

- [] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


src/HealthChecks.Kafka/DependencyInjection/KafkaHealthCheckBuilderExtensions.cs
```diff
@@ -113,4 +113,36 @@ public static IHealthChecksBuilder AddKafka(
...
+    /// <param name="optionFactory">Options to configure Kafka health check.</param>
+    /// <param name="name">The health check name. Optional. If <c>null</c> the type name 'kafka' will be used for the name.</param>
+    /// <param name="failureStatus">
+    /// The <see cref="HealthStatus"/> that should be reported when the health check fails. Optional. If <c>null</c> then
+    /// the default status of <see cref="HealthStatus.Unhealthy"/> will be reported.
+    /// </param>
+    /// <param name="tags">A list of tags that can be used to filter sets of health checks. Optional.</param>
+    /// <param name="timeout">An optional <see cref="TimeSpan"/> representing the timeout of the check.</param>
+    /// <returns>The specified <paramref name="builder"/>.</returns>
+    public static IHealthChecksBuilder AddKafka(
```
adamsitnik, line 130: @mitchdenny would adding this overload solve the problem you have faced in Aspire?

I can see that you have solved it by simply reusing the ctor of the health check and registering your own thing in the DI container:

https://github.com/dotnet/aspire/blob/422d777c8dc7df154d9eea268f0d28a35e673f1b/src/Components/Aspire.Confluent.Kafka/AspireKafkaProducerExtensions.cs#L139-L151

src/HealthChecks.Kafka/DependencyInjection/KafkaHealthCheckBuilderExtensions.cs
```diff
@@ -113,4 +113,36 @@ public static IHealthChecksBuilder AddKafka(
...
+        this IHealthChecksBuilder builder,
+       Func<IServiceProvider, KafkaHealthCheckOptions> optionFactory,
```
adamsitnik, line 132: nit: whitespace

```suggestion
        Func<IServiceProvider, KafkaHealthCheckOptions> optionFactory,

```

adamsitnik (commented):
I would prefer to wait for feedback from Aspire if adding this particular overload is going to solve the problem. If so, we are going to need some new tests (just a few).

Also, I wonder if we should instead add an overload that requires a `IProducer<K, V>` factory .

Something like this:

```cs
public static IHealthChecksBuilder AddKafka<TKey, TValue>(
    Func<ServiceProvider, IProducer<TKey, TValue>>? providerFactory = default,
    Func<Message<TKey, TValue>> messageBuilder,
    string topic = DEFAULT_TOPIC,
    string? name = default,
    TimeSpan? timeout = default)
{
        name ?? NAME,
        sp => new KafkaHealthCheck<TKey, TValue>(providerFactory?.Invoke(sp) ?? sp.GetRequiredService<IProducer<TKey, TValue>>(), messageBuilder, topic),
        failureStatus,
        tags,
        timeout));
}

public class KafkaHealthCheck<TKey, TValue> : IHealthCheck, IDisposable
{
    private readonly IProducer<TKey, TValue> _producer;
    private readonly Func<Message<TKey, TValue>> _messageBuilder;
    private readonly string _topic;

    public KafkaHealthCheck(IProducer<TKey, TValue> producer, Func<Message<TKey, TValue>> messageBuilder, string? topic)
    {
        _producer = producer;
        _messageBuilder = messageBuilder;
        _topic = topic ?? KafkaHealthCheckBuilderExtensions.DEFAULT_TOPIC;
    }

    /// <inheritdoc />
    {
        try
        {
            Message<TKey, TValue> message = _messageBuilder.Invoke();

            var result = await _producer.ProduceAsync(_topic, message, cancellationToken).ConfigureAwait(false);

            if (result.Status == PersistenceStatus.NotPersisted)
            {
                return new HealthCheckResult(context.Registration.FailureStatus, description: $"Message is not persisted or a failure is raised on health check for kafka.");
            }

        }
        catch (Exception ex)
        {
        }
    }

    public void Dispose() => _producer.Dispose();
}
```

This would force the users to register and resolve everything on their own in the DI and we would not need to care about settings.

adamsitnik:
> @adamsitnik, I prefer that overload too but this mean we should remove all other overloads and introduce breaking changes. Should we do this?

This time I am not 100% convinced as the new approach would require the users to always specify `TKey`, `TValue` and the message builder. So it would not be super easy to move to the new version.

I was hoping to get feedback from @mitchdenny but it seems that he just started his winter holiday leave.

@eerhardt what are your thoughts on this? Should we introduce such a breaking change or perhaps just extend the existing ways with a new overload?

src/HealthChecks.Kafka/DependencyInjection/KafkaHealthCheckBuilderExtensions.cs
```diff
@@ -113,4 +113,36 @@ public static IHealthChecksBuilder AddKafka(
...
+        HealthStatus? failureStatus = default,
+        IEnumerable<string>? tags = default,
+        TimeSpan? timeout = default)
+    {
+        Guard.ThrowIfNull(builder);
+        Guard.ThrowIfNull(optionFactory);
+
+        return builder.Add(new HealthCheckRegistration(
+            name ?? NAME,
+            sp => new KafkaHealthCheck(optionFactory.Invoke(sp)),
```
eerhardt, line 143: I think this callback gets invoked every time the health is checked. So the optionFactory would be invoked each time as well.

Alirexaa: Correct.

Alirexaa:
> > @eerhardt what are your thoughts on this? Should we introduce such a breaking change or perhaps just extend the existing ways with a new overload?
> 
> My default opinion is to not introduce a breaking change if you can help it. If anything, we could create a new `IHealthCheck` class, if necessary.

I thought about new class and it seems creating a new class is better approach. I think this new factory method could create memory leak.
Also with new class we can resolve singlton client from factory method and mark older class and methods as obsolated.

Alirexaa:
@adamsitnik Should we add a new `HealthCheck` class?

PR #2380, opened 2025-03-07
feat: added an overload for the AddS3 extension method to take the IS…

steveski:
feat: added an overload for the AddS3 extension method to take the IServiceProvider

Other health checks pass the IServiceProvider into the action/func parameter. This change add that option. I was in a position in my own project where fetching values from my appsettings.json couldn't work easily. The S3 ServiceUrl, access and secret keys needed to passed as well on the options and the only option seemed to be to hard code. Given .NET's HealthCheckRegistration provides the IServiceProvider it made more sense to place the S3HealthCheck instantian in that delegate and pass the SP through to the caller. Adding as an overload prevents old clients from breaking.

None, feature only.

As it's an overload, all existing calling code will continue working.

- [y] Code compiles correctly
- [y] Created/updated tests
- [y] Unit tests passing
- [x] End-to-end tests passing - I only added an optional extension method and I'm not in a position to do end to end testing on the entire suite of health checks you provide

### DIFF
--- a/src/HealthChecks.Aws.S3/DependencyInjection/S3HealthCheckBuilderExtensions.cs
+++ b/src/HealthChecks.Aws.S3/DependencyInjection/S3HealthCheckBuilderExtensions.cs
@@ -41,4 +41,40 @@ public static class S3HealthCheckBuilderExtensions
             tags,
             timeout));
     }
+
+    /// <summary>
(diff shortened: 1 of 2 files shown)


steveski:
dotnet-policy-service agree [company="sole trader"]
----
If God hadn't intended us to have a 3 martini lunch, then why do you think he put all those olive trees in the holy land?

Issue #2364, opened 2025-01-10
AspNetCore.HealthChecks.Rabbitmq leaks connections when using the factory


JanEggers:
**What happened**:
when using 

 ```csharp
  services.AddHealthChecks()
.AddRabbitMQ(factory: c => /* my factory does not matter*/)
```

the healthcheck does not close the connection causing a massive ammount of connections to be opened

**What you expected to happen**:
either one of these:
- the healthcheck should cache the connection after it was initially created and use that one connection for subsequent health checks
- the healthcheck should close the connection after healthcheck is completed if the connection was created with the factory 
- drop the factory support and require the connection to be provided by DI (I would discourage this approach as the connection requires async code to be created which does not play nicely with di as far as im aware and the creation may fail if there is no broker running)

Version 8.0.2 was working fine Version 9.0.0 contains the issue.

propably this change: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/pull/2343

JanEggers:
FYI my solution for the time beeing:

services.AddSingleton<Lazy<Task<IConnection>>>(c =>
    new Lazy<Task<IConnection>>(() => /* my factory does not matter*/);

services.AddHealthChecks()
    .AddRabbitMQ(factory: c => c.GetRequiredService<Lazy<Task<IConnection>>>().Value.Result);

that way I still have access to the service provider while creating the connection (I use it to resolve configuration)

adamsitnik:
Why do you register a `Lazy` factory? The DI container won't create it until it's needed for the first time.

cc @eerhardt

JanEggers:
I guess its not needed then.

kooshan:
This is a really fucked up way to setup a so called FACTORY!!! for health check. Wasted my whole day on finding the source of resource leakage. It looks more like an unintentional bug now being disguised as an intentional clever design. Please do fix this atrocity.

That setup guide in the readme offers no help either.
1) What if I have another service which will use `IConnection` and I don't want to use singleton registration for that?
2) How can I access my IConfiguration for rabbit user/pass when I use the static Lazy to initialize a factory? Should I always hardcode my user/pass in my code?

Wouldn't it be easier to add a `using` at the place where you invoke the factory instead of this nonsense?

VsevolodIO:
Here is the working solution.
```csharp
[...]

PR #2393, opened 2025-06-23
Fixed Connection Leak for RabbitMQ

HesamKashefi:
**What this PR does / why we need it**: The connection leak in RabbitMQ Health Checks made this library unusable

**Which issue(s) this PR fixes**: The connection leak in RabbitMQ

Please reference the issue this PR will close: #2364

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/src/HealthChecks.Rabbitmq/RabbitMQHealthCheck.cs
+++ b/src/HealthChecks.Rabbitmq/RabbitMQHealthCheck.cs
@@ -28,7 +28,7 @@ public class RabbitMQHealthCheck : IHealthCheck
     {
         try
         {
-            var connection = _connection ?? await _factory!(_serviceProvider!).ConfigureAwait(false);
+            await using var connection = _connection ?? await _factory!(_serviceProvider!).ConfigureAwait(false);
 
             await using var model = await connection.CreateChannelAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
 


ThumbGen:
@HesamKashefi Could you please reply with the license/cla so this important PR can move on?

Alirexaa:
@HesamKashefi, can you ensure that the connection is not disposed of in the second check and the next one?
Please add a test about that.

ThumbGen:
Isn't this fix disposing the singleton connection (if configured as per Readme recommendations)? Is the configuration suggested in Readme not addressing the issue? I am not near my dev machine to try it, but seems OK, and makes this PR not necessary.

HesamKashefi:
@dotnet-policy-service agree

HesamKashefi:
> @HesamKashefi, can you ensure that the connection is not disposed of in the second check and the next one? Please add a test about that.

This class uses a factory function to get an instance of the connection.
If you create an instance inside the factory function everything will be OK, but if you provide a factory function that returns a Singleton Instance, then yes it will be disposed and that would be a problem!
But the main problem is that this code seems unusable!

If you take the first approach and create an instance on every call, the instances won't be disposed and the connection leak explodes the server and if you take the second approach and want to return a singleton instance, the connection MUST be created when RabbitMQ is already running and connection will be lost when RabbitMQ goes down! this is unusable!


The third option is to provide a connection (which It's not possible if RabbitMQ is not running)

Am I doing it wrong? I just want to make this work when the RabbitMQ is possibly not available!

HesamKashefi:
> I am not near my dev machine to try it, but seems OK, and makes this PR not necessar

Please check the situation when RabbitMQ is not running.
Please read this issue #2393 .

p.s.

Registering a Singleton like this looks awful! 
Calling `.GetAwaiter().GetResult();` is not a good way to code and creating a singleton connection to a server that might not exists is worse!

```
            builder.Services.AddSingleton(s =>
            {
                var cf = new RabbitMQ.Client.ConnectionFactory
                {
                    HostName = builder.Configuration.GetValue<string>("RabbitMq:Host") ?? "localhost",
                    UserName = builder.Configuration.GetValue<string>("RabbitMq:UserName") ?? "",
                    Password = builder.Configuration.GetValue<string>("RabbitMq:Password") ?? ""
                };
                return cf.CreateConnectionAsync().GetAwaiter().GetResult();
            });
```

ThumbGen:
Just to summarize:

- using an automatically created singleton IConnection is not wanted anymore (this is what v8.x was doing)
- creating a new connection for every health check is not following RabbitMq recommendations https://www.rabbitmq.com/docs/connections#high-connection-churn
- expecting the client application to manage the connection feels weird, as I expect a health checks library to "do what is supposed to do" without me interfering with its internals

🤔

HesamKashefi:
> Just to summarize:
> 
>     * using an automatically created and maintained IConnection is not wanted anymore (this is what v8.x was doing)
> 
>     * creating a new connection for every health check is not following RabbitMq recommendations https://www.rabbitmq.com/docs/connections#high-connection-churn
> 
>     * expecting the client application to manage the connection feels weird, as I expect a health checks library to "do what is supposed to do" without me interfering with its internals, especially forcing me to use .Result
> 
> 
> 🤔

Yes you are right about the fact that creating a new connection is not a good way.
I was just trying to don't touch any code and just fix the problem that I think it's a bug.
Maybe I've a misunderstanding of something here!
But I just couldn't use this code! as I mentioned the problem above!
So instead I'm using this: (just a temporary workaround)
```

            builder.Services.AddHealthChecks()
            .AddCheck("self", () => HealthCheckResult.Healthy())
            .AddRedis("redis")
            .AddNpgSql(builder.Configuration.GetConnectionString("Default")!)
            .AddAsyncCheck("RabbitMQ", async () =>
            {
                try
                {
                    var cf = new RabbitMQ.Client.ConnectionFactory
                    {
                        HostName = builder.Configuration.GetValue<string>("RabbitMq:Host") ?? "localhost",
                        UserName = builder.Configuration.GetValue<string>("RabbitMq:UserName") ?? "",
                        Password = builder.Configuration.GetValue<string>("RabbitMq:Password") ?? ""
                    };
                    await using var connection = await cf.CreateConnectionAsync();
                    if (connection.IsOpen)
                }
                catch (Exception e)
                {
                    LogManager.GetCurrentClassLogger().Error(e);
                }
                return HealthCheckResult.Unhealthy();
            });
```

I think AddRabbitMQ must either get a fix or get completely removed!

PR #2398, opened 2025-07-26
fix: remove manual connect to SurrealDB instance on checkl

Odonno:
There is an issue with the current implementation of the `SurrealDB` healthcheck: calling `Connect` when already connected throws an exception.

Since connection and reconnection are handled internally in the library, there is no need to call the method manually. It will also be called internally within `Health` if the client is not connected.


Alirexaa:
LGTM, thanks

PR #2433, opened 2025-09-17
Fixes ServiceBus PeekMessageAsync Cancellation Token bug

### LINKED ISSUE #2432: Service Bus queue and subscription health check does not respect cancellation token
**What happened**:
When using the PeekMessagesAsync health check on a queue or subscription, the processor doesn't respect the cancellation token passed to the method. This is because internally the AMPQ client uses timeouts instead of cancellation tokens in their retry policies, by default 1 minute (+5 seconds that are always added for some reason)

I would expect the health check to time out when the cancellation token is triggered.

With connection string pointing to an unreachable destination, register a queue or subscription health check with a timeout. This timeout is not respected.

**Source code sample**:
`builder.Services.AddHeathChecks.AddAzureServiceBusQueue(<connstr>, "queue", timeout: TimeSpan.FromSeconds(5));`

This is especially important for production environments with health checks enabled. Those usually implement a strategy of RequestTimeouts and OutputCaching to prevent cascading health probes causing denials of service. The aspnet core request timeout middleware is not aggressive. Meaning that it doesn't kill the request or respond even if the request hasn't finished. It only triggers the HttpContext.RequestAborted cancellation token. If a health probe doesn't respect the cancellation token, the request is not answered within the timeout and the output cache doesn't get set. The healthcheck should register a callback on the cancellation token and respond unhealthy on it, even if the call to PeekMessagesAsync hasn't completed.

**Environment**:
- .NET Core version 9.0
- Healthchecks version 9.0.0
- Operative system: Windows

andradf:
ServiceBusReceiver.PeekMessageAsync does not necessarily respect the cancellation token. See https://github.com/Azure/azure-sdk-for-net/issues/52620. This causes the SB Queue and Subscription health checks to not return. In production environments where request timeouts and output cache for health endpoints are vital to avoiding health check cascading and DDOS attacks, this limitation prevents the usual request timeout and output caching patterns in AspNet Core from operating correctly. 
To get around this, this PR fails the health check if the receiver.PeekMessage does not return before the cancellation token is triggered.

Please reference the issue this PR will close: #2432

- [x] Code compiles correctly (The Dapr project fails due to deprecated dependency warnings, but that issue is already present in master)
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.AzureServiceBus/AzureServiceBusQueueHealthCheck.cs
+++ b/src/HealthChecks.AzureServiceBus/AzureServiceBusQueueHealthCheck.cs
@@ -43,8 +43,15 @@ public class AzureServiceBusQueueHealthCheck : AzureServiceBusHealthCheck<AzureS
                 _queueKey,
                 _ => client.CreateReceiver(Options.QueueName))
                 .ConfigureAwait(false);
-
-            await receiver.PeekMessageAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
+            using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            var cancel_task = Task.Delay(Timeout.Infinite, cts.Token);
+            var peek_task = receiver.PeekMessageAsync(cancellationToken: cancellationToken);
+            var winner = await Task.WhenAny(peek_task, cancel_task).ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


andradf:
@dotnet-policy-service agree

PR #2459, opened 2026-01-12
test queue and topic send using CreateMessageBatchAsync to avoid management grants

### LINKED ISSUE #555: Azure Service Bus Topic hc fails with "Manage,EntityRead claims required for this operation"
The Azure Service bus health checks started to fail after upgrade from version 3.1.1 to version 3.2.1 of AspNetCore.HealthChecks.AzureServiceBus

If I change the policy on the topic from Send, Listen to Manage the exception/unhealthy check is gone and it works as expected.

I would expect it to work the same as version 3.1.1, without the need to change the policy. I don't expect the health check to have manage policy. It should at least be optional.

**Source code sample**:

I am using the health check for topics. 

``` .net
hcBuilder
                .AddAzureServiceBusTopic(
                    configuration[ApplicationConstants.ServiceBusConnectionString],
                    topicName: configuration[ApplicationConstants.ServiceBusTopic],
                    name: "servicebus-check",
                    tags: new string[] { "servicebus" },
                    failureStatus: HealthStatus.Unhealthy);
```

The response is:

``` json
 "servicebus-check": {
            "data": {},
            "description": "Manage,EntityRead claims required for this operation. ......",
            "duration": "00:00:00.3142384",
            "exception": "Manage,EntityRead claims required for this operation.......",
            "status": "Unhealthy"
        }
```

**Environment**:

- .NET Core version: 3.1
- Healthchecks version: AspNetCore.HealthChecks.AzureServiceBus 3.2.1
- Operative system: Windows 10
- Others:

### LINKED ISSUE #727: Modify Azure Service Bus Health Check to no require a connection string with management rights

I noticed that in order for the health check calls `AddAzureServiceBusSubscription()` and `AddAzureServiceBusTopic()` to work properly, we need to submit a connection string with management rights. I think it would be preferable if that was not required and if the health check could still be done even if the connection string only had listen or send rights. 

**Why is this needed**:
Requiring management rights goes against the principles of least privileges I think and it forces me to decide between the convenience of using the library at the risk of giving more access that I want to my app versus only giving it the access it needs but having to write custom health check for the service bus.

maggsimo:
feat: test queue and topic send using CreateMessageBatchAsync to avoi…

This PR introduce a way to test topic or queue using send grant to avoid that we force to provide management roles to implement the hc

Please reference the issue this PR will close: #555
Please reference the issue this PR will close: #727 

i'm leveraging **CreateMessageBatchAsync** since it does a network call to the Service Bus to fetch entity properties, such as MaxMessageSize.

Therefore, it verifies:
- The entity exists
- User connection and permissions allow access (send link)
- User know the maximum allowed message size

It does NOT send messages, **so no DLQ or queue pollution occurs**

Potentially could be possible also to use a probe message to verify that the configuration of the message is valid.

**Sample**
```csharp
  .AddAzureServiceBusQueue(publisherQueueConnectionString,
                          publisherQueueName,
                          configure: (op) =>
                          {
                              op.UsePeekMode = false;
                              op.UseCreateMessageBatchAsyncMode = true;
                          },
                          name: "test queue publish lib",
                          timeout: TimeSpan.FromSeconds(20))
```

```csharp
  .AddAzureServiceBusTopic(publisherTopicConnectionString,
                          publisherTopicName,
                          configure: (op) =>
                          {
                              op.UseCreateMessageBatchAsyncMode = true;
                          },
                          name: "test topic publish lib",
                          timeout: TimeSpan.FromSeconds(20))
```

Yes.
For the queue, since queue can be consumed in both SEND and LISTEN mode, to mirror existing implementation i've set UseCreateMessageBatchAsyncMode = true as default.
This implies that if you want to test only LISTEN, is mandatory to explicitely set _configure_ in _AddAzureServiceBusQueue_ call, to maintein the same behaviour of current implementation,  like:
```csharp
  .AddAzureServiceBusQueue(subscriberQueueConnectionString,
                          subscriberQueueName,
                          configure: (op) =>
                          {
                              op.UsePeekMode = true;
                              op.UseCreateMessageBatchAsyncMode = false;
                          },
                          name: "test queue subscriber",
                          timeout: TimeSpan.FromSeconds(20))
```

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/src/HealthChecks.AzureServiceBus/AzureServiceBusQueueHealthCheck.cs
+++ b/src/HealthChecks.AzureServiceBus/AzureServiceBusQueueHealthCheck.cs
@@ -26,7 +26,11 @@ public class AzureServiceBusQueueHealthCheck : AzureServiceBusHealthCheck<AzureS
         {
             if (Options.UsePeekMode)
                 await CheckWithReceiver().ConfigureAwait(false);
-            else
+
+            if (Options.UseCreateMessageBatchAsyncMode)
+                await CheckWithSender().ConfigureAwait(false);
+
+            if (Options.UsePeekMode is false && Options.UseCreateMessageBatchAsyncMode is false)
                 await CheckWithManagement().ConfigureAwait(false);
 
             return HealthCheckResult.Healthy();
@@ -45,6 +49,17 @@ public class AzureServiceBusQueueHealthCheck : AzureServiceBusHealthCheck<AzureS
                 .ConfigureAwait(false);
 
             await receiver.PeekMessageAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
+        }
+
+        async Task CheckWithSender()
+        {
+            var client = await ClientCache.GetOrAddAsyncDisposableAsync(ConnectionKey, _ => CreateClient()).ConfigureAwait(false);
+            var sender = await ClientCache.GetOrAddAsyncDisposableAsync(
+                _queueKey,
+                _ => client.CreateSender(Options.QueueName))
+                .ConfigureAwait(false);
+
+            await sender.CreateMessageBatchAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
         }
 
         Task CheckWithManagement()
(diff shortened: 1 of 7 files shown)


maggsimo:
@dotnet-policy-service agree

